PR #27, opened 2021-06-12
crontabs: stop killing dnsmasq

PolynomialDivision:
Why should we periodically kill dnsmasq? Stop doing that.


SvenRoederer:
for reference: it's not KILLed, just HUPed to reread the datafiles. see PR https://github.com/freifunk-berlin/firmware/issues/618 message.

PR #59, opened 2021-09-07
more verbose README

Akira25:
Add some documentation on the structure and where to add stuff and so on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,41 +1,196 @@
 # bbb-configs (Ansible Config Management)
 
+## FAQ
+
 ### How to get started
 
-```
+```sh
 pip3 install -r requirements.txt
 ```
 
 ### How to spin up a config run (generate only, output path is /tmp/configs/..:
 
-```
+```sh
 ansible-playbook play.yml
 ```
 
 ### How to spin up a config run and generate images
 
-```
+```sh
 ansible-playbook play.yml --tags image
 ```
 
 ### How to limit image creation to a single location
 
-```
+```sh
 ansible-playbook play.yml --limit example-core
 ansible-playbook play.yml --limit example-*
 ```
 
+The second command doesn't work in `zsh`. If you encounter that problem consider using `bash` instead, please.
+
 ### How to spin up a config run, generate image and flash (Requires IPv6 Connectivity from inside freifunk network, not yet fully working)
 
-```
+```sh
 ansible-playbook play.yml --tags flash
 ```
 
 ### Whats required to bringup a location?
 
-```
-/group_vars : Create location folder
-/host_vars : Create host folder for every openwrt device
-Done!
+1. Create a location folder at `/group_vars/` and fill in at least `general.yml`, `networks.yml` and `owm.yml`.
+2. Create a folder for every OpenWrt device at the location under `/host_vars/`. Paste the `base.yml` in there.
+3. Done!
 
+
+## Variables and structures explained
+
+### group_vars/
+
+`group_vars/` holds tree types of resources: the `all/`-directory, location directories and model files.
+
+#### all/
+
+This directory holds base configuration options that will be inserted into every router configuration. Each device will get this options, like timezones, DNS-servers, the packagefeed-URL and so on.
+
+There are also files for the standard ssh keys and definitions for the wifi profiles.
+
+#### model_files
+
+These files define how bbb-configs needs to handle different hardware models. This example shows a WDR4900:
+```yml
+---
+imagebuilder: "https://downloads.openwrt.org/releases/21.02-SNAPSHOT/targets/mpc85xx/p1010/openwrt-imagebuilder-21.02-SNAPSHOT-mpc85xx-p1010.Linux-x86_64.tar.xz"
+feed_version: "1.2.0-snapshot"
+imagebuilder_disable_signature_check: true      # ignore missing signatures for packages (i.e. selfcompiled)
+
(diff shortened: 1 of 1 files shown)


Akira25:
feel free to comment

README.md
```diff
@@ -1,41 +1,121 @@
...
+#### model_files
+
+These files define how bbb-configs needs to handle different hardware models. This example shows a WDR4900:
+```sh
+---
+imagebuilder: "https://downloads.openwrt.org/releases/21.02-SNAPSHOT/targets/mpc85xx/p1010/openwrt-imagebuilder-21.02-SNAPSHOT-mpc85xx-p1010.Linux-x86_64.tar.xz"
+feed_version: "1.2.0-snapshot"
+imagebuilder_disable_signature_check: true      # don't check the signature-keys when installing packages ??
+
+switch_ports: 6             # What's that for?
```
PolynomialDivision, line 66: number of ethernet ports a device has + cpu

README.md
```diff
@@ -1,41 +1,121 @@
...
+switch_int_port: 0          # What's that for?
```
PolynomialDivision, line 67: cpu port

README.md
```diff
@@ -1,41 +1,121 @@
...
+switch_ignore_ports: []     # What's that for?
```
PolynomialDivision, line 68: sometimes, not all switch ports are usuable. for example the sxtsq 5 ac has "6" ports, but oly cpu port + port 5 is used (1 port device)

README.md
```diff
@@ -1,41 +1,121 @@
...
+
+int_port: eth0              # What's an int-port ??
```
PolynomialDivision, line 70: what swconfig device, e.g. `eth0` or `eth1`

README.md
```diff
@@ -1,41 +1,121 @@
...
+There are also files for the standard ssh keys and definitions for the wifi profiles.
+
+#### model_files
+
+These files define how bbb-configs needs to handle different hardware models. This example shows a WDR4900:
+```sh
+---
+imagebuilder: "https://downloads.openwrt.org/releases/21.02-SNAPSHOT/targets/mpc85xx/p1010/openwrt-imagebuilder-21.02-SNAPSHOT-mpc85xx-p1010.Linux-x86_64.tar.xz"
+feed_version: "1.2.0-snapshot"
+imagebuilder_disable_signature_check: true      # don't check the signature-keys when installing packages ??
```
PolynomialDivision, line 64: @Akira25 Are you checking keys? I think we currently only rely on https when downloading packages. However, nice feature improvement.

PolynomialDivision:
nice thanks for documentation! maybe adding some dsa device would be also nice.

README.md
Akira25: Of course we check the keys on falter. OpenWrt-21.02 imagebuilders don't work without it. Why does ansible deactivate this? Having the hassle with the keys and encryption and not using them for security seems pointless to me.

Akira25:
> nice thanks for documentation! maybe adding some dsa device would be also nice.

If you give me a hint, where to find such a configuration, I could do that.

README.md
spolack: Its not yet implemented
https://github.com/Freifunk-Spalter/bbb-configs/issues/19

spolack:
> If you give me a hint, where to find such a configuration, I could do that.

https://github.com/Freifunk-Spalter/bbb-configs/blob/master/group_vars/model_ubnt_edgerouter_x_sfp.yml

There you can also see how poe ports are being described. On https://github.com/Freifunk-Spalter/bbb-configs/blob/ab19e96532aa6d0ab58cd1425ca144288df5f034/host_vars/koepi-trailer1/base.yml#L8 you can see, how to enable individual POE ports.

Akira25:
@spolack @PolynomialDivision I think this should be ready now.

README.md
```diff
@@ -1,41 +1,196 @@
...
+location_nice: roof top, Example-Street 24, 10573 Berlin        # any string that describes your location. Could be an address or something
+latitude: 52.484948320
+longitude: 13.443380903
+```
+
+##### networks.yml
+
+This file is the most important one of your setup. It defines IP-Addresses, WIFI-Properties, Mesh and so on. Lets have a (shortened) example of a `networks.yml` from the magda-location:
+
+```sh
```
spolack, line 114: That should be yaml i guess. I never use that infostring, not sure if you did this on purpose

README.md
```diff
@@ -1,41 +1,196 @@
...
+---
+
+location: sama                  # locations short name, like def'd in location-directories
+role: corerouter                # devices role. Could either be 'corerouter', 'ap' or 'uplink_gateway'
+target: "ipq40xx/generic"       # target of router model
+model: "avm_fritzbox-7530"      # model name like written in the corresponding file name in group_vars/
+
+wireless_profile: freifunk_default      # activates wifi with freifunk-default-settings on this device
```
spolack, line 174: .. by overriding default wireless profile for corerouters, which is the profile `disable`

spolack (approved):
Thank you for your contribution. very appreciated!

README.md
Akira25: I did it on purpose.vscode complained, that markdown standard requires that.

README.md
Akira25: added in new commit-version

Akira25:
@spolack Threres one approval already. Can we merge this?

spolack:
Sure!

PR #70, opened 2021-09-27
gubener: Add robertfoss ssh key

robertfoss:
Introduce support for ignoring global ssh-keys in location_xxx/ssh-keys.yml
    by defining or assigning a local location ssh-key instead.
    
    Using both the global and local ssh-keys is done by defining the
    'location__ssh_keys__to_merge' variable as follows.
    
    location__ssh_keys__to_merge:
      - comment: XXX
        key: XXX
    
    If only the local ssh-key should be used, define the
    'only_location__ssh_keys__to_merge' variable as follows.
    
    only_location__ssh_keys__to_merge:
      - comment: XXX
        key: ssh-rsa XXX


spolack:
This will not remove other ssh keys for your location. Let me know if this what you want.

robertfoss:
I made the `ssh_keys__to_merge` merge_vars prefix conditional on `only_location__ssh_keys__to_merge` being defined.

Secondly I added a second merge_vars prefix which only merges the 'only_location__ssh_keys__to_merge' prefix to the ssh_keys variable.

spolack (changes requested):
forgot to submit the review. i'm sorry

spolack:
lgtm

PR #78, opened 2021-09-29
ipq40xx: add option to define cpu_settings

PolynomialDivision:
Definining different cpu settings can result in performance increase
and reduction of latency/jitter.

### DIFF
--- a/group_vars/target_ipq40xx_generic
+++ b/group_vars/target_ipq40xx_generic
@@ -5,3 +5,13 @@ target__packages__to_merge:
   - naywatch
 
 multicore: true
+
+sysfs_overrides:
+  - path: /sys/devices/system/cpu/cpu0/cpufreq/scaling_min_freq
+    value: 200000
+  - path: /sys/devices/system/cpu/cpufreq/ondemand/sampling_rate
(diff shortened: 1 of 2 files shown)


robertfoss:
@PolynomialDivision I think this patch looks good and achieves the targeted goals.

@Akira25 What would adding these configuration values to upstream OpenWRT look like? Are 'optimized defaults' maintained for other platforms?

spolack:
> @Akira25 What would adding these configuration values to upstream OpenWRT look like? Are 'optimized defaults' maintained for other platforms?

i think "optimized defaults" heavily depend on and vary with the usecase. Not sure if that belongs upstream.

PolynomialDivision:
Should I go for performance?

robertfoss:
I think `ondemand` is preferable to `performance`, due to thermal reasons.

@Akira25 Do you object to merging this PR?

Akira25:
No, I'm fine with merging this. @spolack's explanation sounds reasonable.

> 
> i think "optimized defaults" heavily depend on and vary with the usecase. Not sure if that belongs upstream.

PR #84, opened 2021-10-30
tp-link-eap225: switch to snapshot and ct driver

PolynomialDivision:
I prefer going to trunk and use the ct driver. The device seems unstable
and experiences crashes.

### DIFF
--- a/group_vars/model_tplink_eap225_outdoor_v1.yml
+++ b/group_vars/model_tplink_eap225_outdoor_v1.yml
@@ -1,12 +1,8 @@
 ---
-imagebuilder: "https://downloads.openwrt.org/releases/21.02-SNAPSHOT/targets/ath79/generic/openwrt-imagebuilder-21.02-SNAPSHOT-ath79-generic.Linux-x86_64.tar.xz"
+imagebuilder: "https://downloads.openwrt.org/snapshots/targets/ath79/generic/openwrt-imagebuilder-ath79-generic.Linux-x86_64.tar.xz"
 feed_version: "1.2.0-snapshot"
 imagebuilder_disable_signature_check: true
 target: ath79/generic
-
-model__packages__to_merge:
-  - "-kmod-ath10k-ct -ath10k-firmware-qca9888-ct"
-  - "kmod-ath10k ath10k-firmware-qca9888"
 
 int_port: eth0
 


group_vars/model_tplink_eap225_outdoor_v1.yml
```diff
@@ -1,13 +1,9 @@
 ---
-imagebuilder: "https://downloads.openwrt.org/releases/21.02-SNAPSHOT/targets/ath79/generic/openwrt-imagebuilder-21.02-SNAPSHOT-ath79-generic.Linux-x86_64.tar.xz"
+imagebuilder: "https://downloads.openwrt.org/snapshots/targets/ath79/generic/openwrt-imagebuilder-ath79-generic.Linux-x86_64.tar.xz"
 feed_version: "1.2.0-snapshot"
```
Akira25, line 3: you should switch that to `master` too, to get the packagefeed thats for snapshot.

Akira25 (changes requested):

group_vars/model_tplink_eap225_outdoor_v1.yml
PolynomialDivision: There is a master feed? Did not know that.

Akira25: Oh, I miss-spelled that one. It's called `snapshot`, not master. But yes, it was there since the very first days.
1.2.0-snapshot is build with the 21.02-SDK, the snapshot feed with the snapshot one. You should defnitely go with the snapshot-feed, if you are going to use snapshot-imagebuilders.

https://firmware.berlin.freifunk.net/feed/snapshot/

PolynomialDivision: thanks I will try!

spolack:
> I prefer going to trunk and use the ct driver. The device seems unstable and experiences crashes.

why?
is -ct driver more stable?

PolynomialDivision:
>  is -ct driver more stable?

It at least can restart radio in terms of crash.

spolack:
Do you have any reference where to find more information about that problem? did you try to get it fixed upstream?

pktpls:
Confirming EAP226-Outdoor occasionally gets a hickup. Wifi connection is up, OLSR neigbor looks good, but somewhere stuff is dropped. OOM-Killer triggers often.

Investigating :-)

Akira25:
@PolynomialDivision @spolack @pktpls Did we get any advancements in this? Otherwise I'd like to close this.

Akira25:
closed by suggestion of @PolynomialDivision in the matrix-chat

PR #85, opened 2021-11-01
treewide: add option for specifying a wifiprefix

PolynomialDivision:
You can now do "whatever.freifunk.net" by setting in your locaiton
    the variable wifiprefix in the wireless_profiles.yml.
    
    Updated locations:
    - colbe15
    - wilgu10

### DIFF
--- a/group_vars/all/wireless_profiles.yml
+++ b/group_vars/all/wireless_profiles.yml
@@ -1,5 +1,7 @@
 ---
 
+wifiprefix: berlin
+
 wireless_profiles:
   - name: disable
     devices:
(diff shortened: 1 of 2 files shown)


Akira25 (changes requested):
I'm against the idea of this PR. Berlin community agreed on having `berlin.freifunk.net` as SSID quite a time ago and has done it ever since.

Having seperate SSIDs per location is really bad. It prevents clients from roaming. In addition users will need to actively look for freifunk-wifis in their reach, instead of just connecting to them.

This PR would introduce a big mess in ther Berlin Freifunk Network.

Akira25:
We had a long discussion on this in our matrix channel. we agreed on this:

On every location, that broadcasts a location-spezific SSID, there must be an additional wifi, that broadcasts on the standard SSID. @PolynomialDivision please change the PR to ensure that, I will approve then. :)

spolack:
@nick are you planning to implement the requested changes? Otherwise i'd like to close that PR

PolynomialDivision:
Lets close.

PR #88, opened 2021-11-05
Add ssh-key for robertfoss to global list

robertfoss:
Add the ssh-key of robertfoss to the global list. Since it is now in the global list, the key no longer needs to be manually added to gubener-core.


spolack (commented):
This would result in all keys getting deployed instead of only yours. is that intended?

PolynomialDivision:
Can you also remove from manstein10?

robertfoss:
@spolack Not exactly intended, but I think it is probably a good idea.

@PolynomialDivision On it.

PolynomialDivision (approved):

PR #90, opened 2021-11-08
.github: add workflow to automatically build hosts

PolynomialDivision:
If you change a location, all hosts of that location will be rebuilt.
If you only change a host, only that host will be rebuilt.


PolynomialDivision:
The second commit is just to show you what is possible.

PR #94, opened 2021-11-13
gubener: Enabled 2+5GHz Mesh networking

robertfoss:


PolynomialDivision:
You want to mesh with the wifi of your core-router?

robertfoss:
Yep! I only have this avm7530 device at my location at this moment. And some neighbors would like to run a ff-falter node until they get internet at home.

PolynomialDivision:
lgtm

PolynomialDivision (approved):

PolynomialDivision:
Just to mention that, maybe it would also be possible to host multiple standlone wifi routers into one location by creating multiple "core routers", like you did. And multiple vlans with (but not required) the same vlan id. However, maybe @spolack sees "one location" as only having one core router. ;)

PolynomialDivision:
Are you allowed to merge yourself?

robertfoss:
No.

robertfoss:
> Are you allowed to merge yourself?

No

robertfoss:
> Just to mention that, maybe it would also be possible to host multiple standlone wifi routers into one location by creating multiple "core routers", like you did. And multiple vlans with (but not required) the same vlan id. However, maybe @spolack sees "one location" as only having one core router. ;)
> 
> Just some theoretical concept.
> 
> You could also directly place the routers into ansible and hope they will survive longer and maintain them here as separate locations. Not sure if you want that.

Just having one router is good for now. At least until I get proper near-field antennas.

PR #101, opened 2021-11-22
uplink_gw: set split_horizon to false

PolynomialDivision:
uplink_gw: set split_horizon to false
    
    With split-horizon some routes contain unnecessary hops via
    the gre-tunnels. Turn it off if the link is a gre-tunnel. Furthermore,
    gre-links are not slow and so no performance bottleneck is given.
    
    https://alioth-lists.debian.net/pipermail/babel-users/2014-June/001631.html


Akira25:
@PolynomialDivision Can we merge this?

spolack:
Please dont apply it without testing that it really makes the difference. That change and explaination is in contrast to my understanding of what the option does.

Akira25:
So should we then close this maybe?

PolynomialDivision:
> So should we then close this maybe?

We don't found the actual cause of the issue in the network. However, it seems to work with split horizon, so I will close.

PR #102, opened 2021-11-25
general.yml: change default contact nickname and mail to info@berlin.*

booo:
As discussed earlier I changed the standard contact information.

### DIFF
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -2,8 +2,8 @@
 zonename: 'Europe/Berlin'
 timezone: 'CET-1CEST,M3.5.0,M10.5.0/3'
 
-contact_nickname: 'please_add_operator_name'
-contact_email: 'please_add_operators_contact'
+contact_nickname: 'Freifunk Berlin Maintenance Team'
+contact_email: 'info@berlin.freifunk.net'
 
 
 dns_servers:
--- a/roles/cfg_openwrt/templates/common/config/freifunk.j2
+++ b/roles/cfg_openwrt/templates/common/config/freifunk.j2
@@ -2,9 +2,7 @@ config public 'contact'
 {% if contact_name is defined %}
 	option name '{{ contact_name }}'
 {% endif %}
-{% if contact_name is not defined or contact_nickname != 'please_add_operator_name' %}
 	option nickname '{{ contact_nickname }}'
-{% endif %}
 	option mail '{{ contact_email }}'
 	option location '{{ location_nice }}'
 


Akira25:
I'd rather prefer to have the direct contact to the location operator there. This was the reason for the last change in a6cee5c9c9fabea7b4118054da755f015054b402. If the operator doesn't want to add its email publicly, he/she can add the email-contact-formular link from the config wizard.

In the mentioned commit you can find some template-code too, which would get broken by this commit.

group_vars/all/general.yml
```diff
@@ -35,4 +35,4 @@ sysupgrade_preserve_custom_files:
   - /etc/wireguard/wg.key
   - /root/
 
-image_search_pattern: "*-sysupgrade.*"
\ No newline at end of file
+image_search_pattern: "*-sysupgrade.*"
```
Akira25, line 38: What is this line for? Did it maybe accidentially get into this commmit?

PolynomialDivision:
@Akira25 We talk here about default values. It is just the "fallback" and not the default. Users should specify contact information in their location. :)

group_vars/all/general.yml
PolynomialDivision: Look here:
https://github.com/Freifunk-Spalter/bbb-configs/commit/49b41b4b4813315bb7cfc9587cd3ffe551f53f31

If you see the diff, philipps editor or whatever just added the newline at the end.

Akira25:
> @Akira25 We talk here about default values. It is just the "fallback" and not the default. Users should specify contact information in their location. :)

I know that. But in the past it hast shown, that most users, especially @freifunkoperator, rely on those default-values. To set an incentive to set the right maintainer contact, We've made the last change to this file, which was approved by @spolack.

To get the people setting right values, the standard value shouldn't make sense.

PolynomialDivision:
Locations with contact information:
- k9
- kiehlufer
- fsev
- gubener
- huette

My opinion is, if all our ssh keys are on it, the email should go to all of us. I would prefer that and it helps if a maintainer becomes unresponsive. Lets share work and knowledge. ;)

Akira25:
> Locations with contact information:
> 
> * k9
> * kiehlufer
> * fsev
> * gubener
> * huette
> 
in short: every location, that doesn't get maintained by you. ;-)

> My opinion is, if all our ssh keys are on it, the email should go to all of us. I would prefer that and it helps if a maintainer becomes unresponsive. Lets share work and knowledge. ;)

Maybe we can add the info-address as second e-mail-address. Ever since the bbb-net started, there was for every location a maintainer. If there is a person, that mainly maintains a location, we should mention this! If that is subject to change, it's really easy, to change that. There is no need to make maintainers "unavailable" by hiding them behind a general e-mail address, which they even don't get the mails from.

booo:
The bbb installations are mostly community "owned" and maintained. There are specific maintainers or people that feel extra responsible for an installation and they can replace the contact if they like. As default a general mail address is good because we maintain the installations as a team and the team/community should be reachable by this mail.

We can add our mailinglist too or add a paragraph to the other fields in the contact site. Maybe a link to the wiki would be nice too.

booo:
Maybe we should discuss this in a meeting and find some consensus for a minimal standard.

spolack:
I'd like to have the mailing list as default. Its better to have the list for community-owned locatinos instead of dead contacts from inactive persons.

For all "@home" locations we should enforce setting a contact that by using review process.

booo:
Problem with the mailinglist is that you have to register first to be able to write to the mailinglist.

By the way we can update the info@ list at some point.

Akira25:
>For all "@home" locations we should enforce setting a contact that by using review process.

Lets do it that way. I've adjusted the template accordingly.

PR #103, opened 2021-11-25
WIP: flughafen: add initial configuration for flughafen

booo:
Still needs a IPv6 prefix, snmp configuration and DFS magic.

Could someone please review my first interaction with this tool. Thanks a lot for your time!

### DIFF
--- a/host_vars/flughafen-core/base.yml
+++ b/host_vars/flughafen-core/base.yml
@@ -1,0 +1,7 @@
+---
+
+location: flughafen
+role: corerouter
+model: "ubnt_edgerouter-x"
+
+wireless_profile: freifunk_default
--- a/group_vars/location_flughafen/networks.yml
+++ b/group_vars/location_flughafen/networks.yml
@@ -1,0 +1,38 @@
+---
+# ipv6_prefix: "::/56" # TODO
+
+# Disable noping as we use dhcp and static assignments
+# both in mgmt network
+dhcp_no_ping: false
+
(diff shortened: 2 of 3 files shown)


host_vars/flughafen-core/base.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+location: fluehafen
```
PolynomialDivision, line 3: fluehafen?

PolynomialDivision:
Please have a look here, we agreed on vlan numbering:
https://github.com/Freifunk-Spalter/bbb-configs/issues/3#issuecomment-889040037

host_vars/flughafen-core/base.yml
booo: Should be floehhafen instead right?

booo:
> Please have a look here, we agreed on vlan numbering: [#3 (comment)](https://github.com/Freifunk-Spalter/bbb-configs/issues/3#issuecomment-889040037)

I don't want to change the vlan ids. Sites with a few more VLANs will run into problems with the proposed scheme anyway.

Akira25:
@PolynomialDivision Why is there already a location flughafen on masterbranch?

PolynomialDivision:
@booo said that we should finalize the flughafen because he has no time.

booo:
Did you use the configuration from here?

Why has flughafen now a totally different IP configuration?

PR #104, opened 2021-11-25
templates: add dummy host to skip static dhcp leases

PolynomialDivision:
templates: add dummy host to skip static dhcp leases
    
    Sometimes we have static assignments. Add them as dummy hostnames to
    dnsmasq.


spolack:
As discussed privately, please implement configuring static dhcp-leases based on the assignments. You can provide the mac address similar to how we set the wifi parameters for individual devices. If no mac is provided, default to a bogus one.

PolynomialDivision:
Seems to work:
image

spolack:
Please also implement defining mac addresses, in order to complete that topic

spolack:
go!

spolack (approved):

PR #107, opened 2021-11-30
uplink: add dnsqmasq to uplink gateways

PolynomialDivision:
Dnsmasq is important to resolve olsr addresses. Add it to uplink gateways.

### DIFF
--- a/group_vars/role_uplink_gateway/imageprofile.yml
+++ b/group_vars/role_uplink_gateway/imageprofile.yml
@@ -38,5 +38,4 @@ role_uplink_gw__packages__to_merge:
   - samplicator
 
 role_uplink_gw__disabled_services__to_merge:
-  - "dnsmasq"
   - "odhcpd"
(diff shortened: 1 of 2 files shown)


PolynomialDivision:
@pmelange that should fix everything

pmelange:
I think that you should also have ```list notinterface {{ uplink['ifname'] }}``` in the ```config dnsmasq``` section.  This will prevent dnsmasq from being a DNS server on the public port.

pmelange:
> I think that you should also have `list notinterface {{ uplink['ifname'] }}` in the `config dnsmasq` section. This will prevent dnsmasq from being a DNS server on the public port.

By the way, is ```uplink['ifname*]``` the device or the interface defined in /etc/config/network?  Because it needs to be the device which is listed in the dhcp config.

pmelange:
for example, in l105-gw
```
config interface 'uplink'
        option ifname 'eth0'
        option proto 'static'
        option ipaddr '77.87.49.8/28'
        option gateway '77.87.49.1'
```

needs to be like this in the dhcp config
```
list notinterface 'eth0'
```

PR #109, opened 2021-12-04
strom-gw: add strom-gw

freifunkoperator:


Akira25 (changes requested):
I addition to the normal owm.yml in groupvars/location, theres a `host_vars/strom-gw/owm.yml` which looks like an error to me. Please end every file with a newline. If you use VSCodium, you might configure it to add the final newline on storing the file.

Final newlines are missing in `host_vars/strom-gw/base.yml`

spolack:
@freifunkoperator can you please update that PR? otherwise prepare for close

PolynomialDivision:
Can be closed due to missing ability to enter place.

PR #112, opened 2021-12-09
hello k-fetisch

FFHener:
Waiting for commit of model_tplink_archer-c7-v5 to be added


FFHener:
The SetUp will probably change. I wait for feedback of the K-Fetisch  (will probably take some time) and will make another PR as soon as it is more clear on whats needed.

Issue #17, opened 2021-03-21
client isolation/nftables races with DFS CAC


spolack:
```
Sun Mar 21 20:22:40 2021 daemon.notice hostapd: wlan5-ff: interface state HT_SCAN->DFS
Sun Mar 21 20:22:40 2021 daemon.notice hostapd: wlan5-ff: DFS-CAC-START freq=5260 chan=52 sec_chan=1, width=1, seg0=58, seg1=0, cac_time=60s


Sun Mar 21 20:22:47 2021 daemon.notice procd: /etc/rc.d/S95done: /etc/config/nftables.conf:16:22-32: Error: Interface does not exist
Sun Mar 21 20:22:47 2021 daemon.notice procd: /etc/rc.d/S95done: 		iif != { wlan5-ff, wlan5-ffowe, wlan2-ff, wlan2-ffowe} accept
Sun Mar 21 20:22:47 2021 daemon.notice procd: /etc/rc.d/S95done: 		```

spolack:
should be fixed with 826be8da57753f4f89a5c992b567c7f552589ae

@FFHener would be great if you could test and verify client_isolation across multiple access points is still working properly.

FFHener:
I will check it and report here

spolack:
its fixed for long time

PR #114, opened 2021-12-14
ssh-key added and new Channel Assignment

FFHener:
I added a ssh-key and Changed the channel of AP4 to a non DFS-channel because of #17

### DIFF
--- a/group_vars/location_fesev/networks.yml
+++ b/group_vars/location_fesev/networks.yml
@@ -35,9 +35,10 @@ networks:
       fesev-core: 1
 
 location__channel_assignments_11a_standard__to_merge:
-  fesev-ap1: 36-40
-  fesev-ap2: 40-40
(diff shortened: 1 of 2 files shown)


spolack:
I noticed there are only two channels available for 40MHz Bandwith in non-dfs range:

40 (is choosen fallback if using 36 with 40MHz bw)
48 (is choosen if using Channel 44 with 40MHZ bw)

I'd propose to assign them in a A B A B scheme, depending on the distance of the APs

FFHener:
Do i have to use Option 1 or 2 in the networks-file?
1)
A: 36-40
B: 44-40
2)
A: 40-40
B: 48-40

spolack:
Both should be fine.

FFHener:
Channels now shoud be good and ready to merge

PR #124, opened 2022-01-06
treewide: prepare meshing via wireguard tunnel

PolynomialDivision:
- add dhcp interface role called uplink
- add ability to specify routing tables for mesh links
- create routing tables
- set policies


spolack:
i think this is obsolete now. Please reopen if not

PR #130, opened 2022-01-16
Fuewa/lebenshilfe

Akira25:
This adds the Location at Lebenshilfe Oder-Spree in Fürstenwalde.


group_vars/location_fffw_lebenshilfe/networks.yml
```diff
@@ -0,0 +1,82 @@
...
+  - vid: 40
+    role: dhcp
+    prefix: 10.30.99.0/25
+    #ipv6_subprefix: 0
+    assignments:
+      fffw-lebenshilfe-core: 1
+
+location__channel_assignments_11g_standard__to_merge:
+  # ap-name: Channel  Bandwidth  tx-Power
+  fffw-lebenshilfe-nno-ap-2ghz: 1-20-15
```
spolack, line 77: is there a reason why you set txpower explicit?

host_vars/fffw-lebenshilfe-core/base.yml
```diff
@@ -0,0 +1,8 @@
+---
+location: fffw-lebenshilfe
+role: corerouter
+target: "ipq40xx/generic"
```
spolack, line 4: its not necessary anymore to set the target per host. Nowadays this mapping went to model group_vars. e.g. https://github.com/freifunk-berlin/bbb-configs/blob/master/group_vars/model_avm_fritzbox_4040.yml#L5

.. It works by using 2 inventory stages. First one creating the group assignments based on host_vars, and a second one which can create group assignments of group_vars which were referenced in the first run.
https://github.com/freifunk-berlin/bbb-configs/tree/master/inventory

host_vars/fffw-lebenshilfe-core/base.yml
```diff
@@ -0,0 +1,8 @@
...
+model: "avm_fritzbox-7530"
+
+wireless_profile: freifunk_fw
```
spolack, line 7: you could consider setting the wireless_profile within your location group_vars. saving some redundant configuration.

we can also think of introducing a group per community. therefore you could override the wireless_profile for all hosts belonging to a community.

host_vars/fffw-lebenshilfe-nno-ap-2ghz/base.yml
```diff
@@ -0,0 +1,11 @@
...
+location: fffw-lebenshilfe
+role: ap
+target: "ath79/generic"
+model: "mikrotik_sxtsq-2-lite"
+
+wireless_profile: freifunk_fw
+
+mac_override:
+  eth0: cc:2d:e0:9c:4d:58
+...
```
spolack, line 11: ?

spolack (changes requested):

group_vars/location_fffw_lebenshilfe/networks.yml
Akira25: No, there isn't. i just copied that from the documentation, which was derived from kiehlufer-core, I think... Can I omit the tx-power and it gets set automatically then?

host_vars/fffw-lebenshilfe-core/base.yml
Akira25: Good to know. Hopefully I will get some time and adjust the examples in the documentation accordingly.

host_vars/fffw-lebenshilfe-core/base.yml
Akira25: The latter solution sounds really attactive to me. For the sake of getting the site online fast, I'd go for the first solution for now. :)

host_vars/fffw-lebenshilfe-nno-ap-2ghz/base.yml
Akira25: I've saw that configuration on a couple of locations using the Mikrotik-SXTsq5ac-antennas. I thought that it needs to be that way. Does OpenWrt get the correct MAC-Address on those devices by itself?

spolack: No the comment is because of L11

Akira25: oh, that one. I 've read somewhere that it's not strictly necessary, but good practice to end a YAML-File with three dots. Like it was begun with three dashes.

spolack:
never seen that before, but you might be right.

group_vars/location_fffw_lebenshilfe/networks.yml
Akira25: tx-power removed in new version.

host_vars/fffw-lebenshilfe-core/base.yml
Akira25: target in host-vars removed.

host_vars/fffw-lebenshilfe-core/base.yml
Akira25: As I had difficulties at defining the wifi profile in group-vars, I left them in host-vars for now.

host_vars/fffw-lebenshilfe-nno-ap-2ghz/base.yml
Akira25: I think that one has been resolved.

spolack (approved):

Akira25:
> Still you could consider basing off `add_tunnelmanager` and enable IPv6. Just choose some District, the routing really doesnt care. And wont in the future, until we restructure IPv6 addressing from scratch

I think, we will postpone IPv6 a bit in Fürstenwalde. For IPv4, we've allocated our own block in the intercity-vpn repository. We didn't speak on IPv6 for, now, but I will ask the other community members of fuerstenwalde what they like to do.

Anyway I think we will add that then. :)

PR #131, opened 2022-01-17
uplink-gateways: continue adding server side wireguard meshing

PolynomialDivision:
Previous commits already add basic wireguard support. However, we still
need more functionality on the server side. This commit finishes this
attempt.

We add IPv4 wireguard meshing and add the "wireguard_base_v4prefix" to
all missing gateways. Ohlauer and Saarbruecker already had those
prefixes. We use OLSR for meshing on IPv4.

Due to the need to change the config file frequently with adding and
removing new WireGuard tunnels, we strain the flash. On boot, we symlink
those config files to /var/run/ like done with other config files to
reduce the load on the flash rom.


group_vars/role_uplink_gateway/general.yml
```diff
@@ -24,11 +24,18 @@ peer_asn: 25291
...
 wireguard_port_end: 52820
 wireguard_wg_key: '/etc/wireguard/wg.key'
 wireguard_wg_pub: '/etc/wireguard/wg.pub'
-
+# Use temporary keys:
+# Generate for each new connection a private and public key in the form of
+# wg_[PORT].key wg_[PORT].pub
+# in /tmp/run/wgserver.
+# If it is set wireguard_wg_key and wireguard_wg_pub are not used.
+wireguard_wg_tmp_key: 1
```
spolack, line 38: Why is that an integer? Thought its boolean.

spolack: Furthermore, is that still nescessary if you implement proper error handling? I'd vote against tbh.

host_vars/ak36-gw/gateway.yml
```diff
@@ -77,6 +77,7 @@ tunnel_metric: 1024
 # Tunnel metric 1024 as most internet uplinks will hardly reach 40MBit/s
 
 wireguard_base_prefix: 2001:bf7:750:4002::/64
+wireguard_base_v4prefix: 10.31.145.0/24
```
spolack, line 80: I dislike v4prefix.

Maybe use `wireguard_base_prefix_ipv4/ipv6`

roles/cfg_openwrt/templates/common/rc.local.j2
```diff
@@ -21,3 +21,12 @@ sed -i '/^ \/etc/d' /usr/lib/opkg/status
 {% for override in sysfs_overrides | default([]) %}
 echo {{ override['value'] }} > {{ override['path'] }}
 {% endfor%}
+
+{% if role == 'uplink_gateway' %}
```
spolack, line 25: Cant you ship that cronjob with the package? Furthermore i find it more appropriate to cleanup more frequently that once a day. What about every 10min?

PolynomialDivision:
@spolack I addressed all changes, however I would really favor the wireguard tmp key solution.

PR #134, opened 2022-01-30
kiehlufer: several changes 2022

Koltonowski:


PolynomialDivision:
lgtm

PolynomialDivision:
The CI fail has nothing to do with the PR. That is why merging
```

MSG:

The task includes an option with an undefined variable. The error was: https://downloads.openwrt.org/{{ 'snapshots' if openwrt_version == 'snapshot' else 'releases/' ~ openwrt_version }}/targets/{{ target }}/openwrt-imagebuilder-{{ openwrt_version ~ '-' if openwrt_version != 'snapshot' }}{{ target|replace('/','-') }}.Linux-x86_64.tar.xz: the inline if-expression on line 1 evaluated to false and no else section was defined.

The error appears to be in '/home/runner/work/bbb-configs/bbb-configs/roles/cfg_openwrt/tasks/imagebuilder.yml': line 8, column 3, but may
be elsewhere in the file depending on the exact syntax problem.
```

PR #137, opened 2022-02-08
template: Throw error if network name is too long

robertfoss:
Due to kernel limitations, the maximum length for a network is 15
characters.

Cause a build failure if the network name is longer than 15 characters,
and output a helpful error message.


Akira25:
Looks good to me.

PR #144, opened 2022-02-25
koepi: allow inbound ipv6

freifunkoperator:


spolack:
Do you really want this and you are aware of the implications exposing the whole location (including management) to the internet?

spolack:
@freifunkoperator please respond or close

PR #146, opened 2022-02-25
rhnk: add olsr penalty to f2a

PolynomialDivision:


PolynomialDivision:
Something failed, but it does not have anything to do with the change:
```
MSG:

The task includes an option with an undefined variable. The error was: [https://downloads.openwrt.org/{{](https://downloads.openwrt.org/%7B%7B) 'snapshots' if openwrt_version == 'snapshot' else 'releases/' ~ openwrt_version }}/targets/{{ target }}/openwrt-imagebuilder-{{ openwrt_version ~ '-' if openwrt_version != 'snapshot' }}{{ target|replace('/','-') }}.Linux-x86_64.tar.xz: the inline if-expression on line 1 evaluated to false and no else section was defined.

The error appears to be in '/home/runner/work/bbb-configs/bbb-configs/roles/cfg_openwrt/tasks/imagebuilder.yml': line 8, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: set destination path for imagebuilder
  ^ here
 ```

File: group_vars/location_rhnk/networks.yml [hunk 1]
```diff
@@ -43,6 +43,8 @@ networks:
     name: mesh_fhain
     prefix: 10.230.3.17/32
     ipv6_subprefix: -17
+    mesh_metric_lqm:
+    - default 0.2 # make sure emma is uplink and not f2a
```
spolack, line 47: That interface is called mesh_fhain. Is f2a the only link? Otherwise please only apply that lqm for the peer ip.

PolynomialDivision: I am not completely sure what is the plan. I guess if it is no ptp link it would be called, `mesh_n` or something like that. However, maybe @a22sc or @booo , or @Koltonowski. I updatet that now to only apply the penalty to f2a.

PR #153, opened 2022-03-05
template: Fix bad syntax from length check

robertfoss:
This commit fixes an issue introduced when https://github.com/freifunk-berlin/bbb-configs/commit/2e432f6286c6be7ed287b92a977d85a437f095d5 added length checking.

```
config wifi-iface 'radio0_if2'
	option device 'radio0'
	option ifname 'wlan5-mesh'
	option mode 'mesh'
	option mesh_id 'Mesh-Freifunk-Berlin'
	option mesh_fwding '0'
	option mcast_rate '12000'
        Mesh name ok
	option network 'mesh_11s_no5'
```

'Mesh name ok' is not mean to be in the above snippet, and causes a parsing issue. 

This commit fixes this syntax error, while keeping the length checking.


PolynomialDivision:
lgtm. thanks! :)

PR #157, opened 2022-03-14
gub37: Grow IP-space to make space for gub37-zwingli

robertfoss:
The old IP space is too small, switch to a new bigger one.

```
IPv4 10.31.157.0/25
IPv6 2001:bf7:830:a700::/56
```

### DIFF
--- a/group_vars/location_gub37/networks.yml
+++ b/group_vars/location_gub37/networks.yml
@@ -1,20 +1,31 @@
 ---
-ipv6_prefix: "2001:bf7:830:4c00::/56"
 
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
I would have put my mgmt to the "0". Otherwise, it looks good. :)

robertfoss:
Yeah, that would have made sense too. I think I'd rather keep it this way than re-configure all of the devices.

robertfoss:
Just pushed fix that marks dhcp vlan as tagged.

PR #158, opened 2022-03-14
gub37: Switch out Hof Nord 5GHz radio

robertfoss:
Switch to a Mikrotik LHG 5 AC radio, which runs the same firmware as SXTsq 5 AC.

### DIFF
--- a/host_vars/gub37-hof-n-5/base.yml
+++ b/host_vars/gub37-hof-n-5/base.yml
@@ -2,6 +2,4 @@
 
 location: gub37
 role: ap
-model: "ubnt_nanostation_loco_m5_xw"
-
-poe_on: []
+model: "mikrotik_sxtsq-5-ac"


host_vars/gub37-hof-n-5/base.yml
```diff
@@ -2,6 +2,6 @@
 
 location: gub37
 role: ap
-model: "ubnt_nanostation_loco_m5_xw"
+model: "mikrotik_sxtsq-5-ac"
 
-poe_on: []
\ No newline at end of file
+poe_on: []
```
PolynomialDivision, line 7: you can delete that line. it has no poe?

PolynomialDivision, line 7: you can delete that line. it has no poe?

robertfoss: Nice catch! Just pushed  a fix.

PR #161, opened 2022-03-17
role_ap: fix remove of firewall

PolynomialDivision:
Firewall is still included in the ap-images because it is a depdency of
luci-app-firewall. Remove also the luci app.


PolynomialDivision:
@spolack has a better solution.

PR #167, opened 2022-03-21
hello beuth

PolynomialDivision:


group_vars/location_beuth/networks.yml
```diff
@@ -0,0 +1,94 @@
...
+    name: mesh_west
+    prefix: 10.230.23.158/32
+    ipv6_subprefix: -28
+
+  - vid: 104
+    role: dhcp
+    prefix: 10.230.23.152/29
+    ipv6_subprefix: 0
+    untagged: true
+    inbound_filtering: true
```
Akira25, line 70: I don't think that we need inbound filtering on beuth. We agreed on doing that in house projects, but not on public sites.

Akira25:
looks good to me.

PR #173, opened 2022-03-23
roles: rework rendering of /etc

pktpls:
Note: with this change, symlinks are only allowed for files, not directories

The motivation for this change is to avoid warning output due to
globbing on non-existing directories. We now use Ansible's filetree
module and:

- for each directory, create a corresponding directory in target /etc
- for stuff with a .j2 name, render it as a template and write it to /etc
- for other files and symlinks, copy it over to /etc.


roles/cfg_openwrt/tasks/render_etc.yml
```diff
@@ -0,0 +1,29 @@
...
+  tags: always
+  file:
+    path: "{{ config_path }}/etc/{{ item.path }}"
+    state: directory
+  with_filetree: "{{ role_path }}/files/{{ role }}"
+  when: item.state == 'directory'
+
+- name: Render /etc templates
+  tags: always
+  template: src={{ item.root }}/{{ item.path }} dest={{ config_path }}/etc/{{ item.path | regex_replace('\.j2$','') }}
```
pktpls, line 21: Using `{{item.root}}/{{item.path}}` instead of just `{{item.src}}` here because if item is a symlink, `item.src` contains the path to the symlink's target, which for some reason is not what we need here

robertfoss:
Thanks for having a look at this.

This looks like a nice fix to me, I'm not seeing any issues with it.

pktpls:
Not sure what's up with the check build, is it hanging, is it cancelled?

I checked locally that we end up with exactly the same /etc output.

roles/cfg_openwrt/tasks/render_etc.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+- name: Create /etc directories part one
```
spolack, line 3: would be great have a better naming than part one and part two. e.g. (static files, templated files)

roles/cfg_openwrt/tasks/render_etc.yml
```diff
@@ -0,0 +1,29 @@
...
+  tags: always
+  file:
+    path: "{{ config_path }}/etc/{{ item.path }}"
+    state: directory
+  with_filetree: "{{ role_path }}/templates/{{ role }}"
```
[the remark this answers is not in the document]
pktpls: Ah yes, had no clue that `with_` syntax is deprecated now :+1:

pktpls:
Thanks for the comments (resolved), I also added some documentation as a cherry on top :heart: 

Again no idea what's up with the check. It ran for about 30 minutes, now it shows as both passed and skipped here, and as both running and failed in the Checks tab. Even got an email with subject "failed" and content "all tests were skipped" :sunglasses:

PolynomialDivision:
That is a bug in the CI. It tries to find the changed location and want to build them again, however it is just empty, so it tries to build everything. That needs to be fixed.

```
Run CHANGED_HOSTS=$(git diff --diff-filter=d --name-only origin/$BRANCH | grep host_vars/ | awk -F/ '{print $2}' | sed -z 's/\n/,/g')
Building for 
Building for 
```

spolack:
I like! :)

PR #177, opened 2022-03-24
beuth: rename to bht

PolynomialDivision:
The Beuth was renamed to Berliner Hochschule für Technik (BHT). Let's
also rename the location.


Noki (approved):
looks good.

PR #189, opened 2022-04-05
Gruni73 channels

robertfoss:
Dodge busy 2ghz channel. Avoid 5ghz channels that overlap with uplink radios.


Akira25 (approved):
Looks good to me.

PR #193, opened 2022-04-18
Maintenance huette & huette15

pktpls:
- add @pktpls ssh key
- codify our wifi settings
- remove mentions of huette15-vm


Akira25:
@pktpls the CI failed on huette-vm. Does this happen on your machine too?

pktpls:
huette-vm is supposed to be removed with the last commit

PolynomialDivision:
> @pktpls the CI failed on huette-vm. Does this happen on your machine too?

That is a failure of the CI. Can not handle removing of hosts. :S

Akira25:
If the CI fails regularily on correct PRs, we don't need it, in my opinion. Its just a useless waste of computing power then. :(

PR #200, opened 2022-05-01
Switch ar40xx devices to snapshot

robertfoss:
In order to use the fixed ar40xx driver, switch to OpenWRT Snapshot.

https://github.com/openwrt/openwrt/commit/e9431a8335658fd8bcb1c01b3c7e59bf0401196d


PolynomialDivision:
Should be unnecessary since the commit is already picked into 21.02:
https://github.com/openwrt/openwrt/commit/e9431a8335658fd8bcb1c01b3c7e59bf0401196d

Soon the buildbot should rebuild the 21.02-snapshot images.

PR #202, opened 2022-05-04
use comments from yml file for generation of authorized_keys

Noki:
This replaces the comment from the original ssh-key with the comment of the yml file, which is the name of the user. This labels keys without a comment with a proper username so it can easily be identified on the router and removes useless information like machine names from the file. The original comment can still be kept in the git repository, where it might be useful to the owner of the key to identify his key.


Akira25 (approved):
Looks good to me and worked fine on my machine.

PR #205, opened 2022-05-11
corerouter/gateway: enable ntp server

PolynomialDivision:
On some sites the airos devices are configured to get the time from
the core router. Enable the server on core-router and uplink-gateway.


Akira25:
looks good to me.

PR #206, opened 2022-05-11
corerouter: add domain search option for olsr domains

PolynomialDivision:
Let the dhcp servers pass name service configuration to DHCP clients
about olsr domains. This will result in:

search olsr
nameserver [IP]

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
+++ b/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
@@ -27,6 +27,7 @@ config dhcp 'dhcp_{{ name }}'
 	option leasetime '5m'
 	option start '2'
 	option limit '{{ network['prefix'] | ipaddr('size') -3 }}'
+	list dhcp_option '119,olsr'
     {% else %}
 	option dhcpv4 'disabled'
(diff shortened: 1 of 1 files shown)


spolack:
why?

Noki:
From what I understand the functionality of this DNS option should be as follows:

> When looking up a bare name in DNS, the network stack will add the search domains to it to form fully qualified domain names, and look up those as well.[[8]](https://en.wikipedia.org/wiki/Search_domain#cite_note-a_support_ml_dns-8) For example, if the domain search list contains "wikipedia.org", typing "en" in the browser will direct the user to "en.wikipedia.org".
Source: https://en.wikipedia.org/wiki/Search_domain

So this will allow all systems that support this to use e.g. `ping example` instead of `ping example.olsr` which is pretty cool.

I think this is a really nice improvement.

Noki (approved):
looks good to me

Akira25 (approved):
Looks good to me. This is implemented in standard-falter too. So it might be a good idea, to keep the whole thing consistently and activate it on bbb-config-nodes too.

spolack:
Please make it at least configurable per network.

PR #209, opened 2022-05-15
Gateways: add ssh-key for dpaufler

Akira25:
The gateways which are sponsored by the förderverein schould be
accessible by it.


spolack:
I'd propose to keep it simple and give him access to every gateway. This can be done in
https://github.com/freifunk-berlin/bbb-configs/tree/master/group_vars/role_uplink_gateway

spolack:
Yep. please change __location to __role. It just indicates, where this variable is from, that is helpful for debugging later.

Akira25:
> Yep. please change __location to __role. It just indicates, where this variable is from, that is helpful for debugging later.

Done.

PR #211, opened 2022-05-21
gruni73: Add Titanium sector antennas

robertfoss:


spolack (changes requested):
Please adjust the TX power to not violate law here. (1-3db cable loss might be reasonable tolerance due to cable loss etc.)

robertfoss:
Thanks @spolack, added tx power levels.

robertfoss:
Adjusted & commented tx power calculation.

PR #220, opened 2022-06-01
Add kts13-core & kts13-ap1

pktpls:


host_vars/kts13-core/base.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+location: kts13
+role: corerouter
+model: "avm_fritzbox-7530"
+
+wireless_profile: disable
```
PolynomialDivision, line 7: I think this is unnecessary. Default is disabled as far as I know.

PR #222, opened 2022-06-02
group_vars/role_uplink_gateway/general.yml: add scherer8-wetter inbound filter

pmelange:
add IP 2001:bf7:750:2e00::10 on port 80 to inbound filter to allow
global access to scherer8-wetter.olsr


PolynomialDivision:
Rolled out diff. You should be able to acces your weather service.

PR #223, opened 2022-06-02
all: remove plugins that scrape mac addresses 

PolynomialDivision:
The plugin scrape the mac addresses of client stations. Remove them.

I am sure there is a way to disable the scraping of mac addresses. However, that needs additional coding. Remove them until it is fixed.

### DIFF
--- a/group_vars/all/imageprofile.yml
+++ b/group_vars/all/imageprofile.yml
@@ -40,13 +40,14 @@ all_luci_base__packages__to_merge:
 all_prompoc__packages__to_merge:
   - curl
   - prometheus-node-exporter-lua
-  - prometheus-node-exporter-lua-hostapd_stations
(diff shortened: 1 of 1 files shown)


spolack:
Please provide a more meaningful commit message and maybe consider reverting all prometheus stuffy for now. it was not properly built, was just a test. We can think of how to reimplement it in a proper way.

Noki:
I think removing only the packages as in this commit is a good compromise for now as it will also solve the problem that error messages from the station exporters pollute the log. Please add some more detail to the commit message and merge this. I really want to keep the prometheus exporters as it makes monitoring the devices so much easier.

Noki (approved):
lgtm

PolynomialDivision:
Replaced by: https://github.com/freifunk-berlin/bbb-configs/pull/266

PR #225, opened 2022-06-08
hds-core: update mesh IP's

pmelange:


Akira25:
Looks good to me. @pmelange explained to me via dm, that we can have double of the adresses on hdm with this change, which is needed as we went out out free adresses there.

Akira25 (approved):

PR #228, opened 2022-06-08
babel: rework metrik; use filter, leave rxcost vanilla

spolack:
With the now enabled link quality estimation, we should leave the rxcost untouched for now, as this is the base which will be multiplied if packetloss is detected. 

Our metric is now explicitly added to the calculated metric.


PolynomialDivision:
LGTM. Run-tested Belkin RT3200 OpenWrt Master. Thanks!

PR #229, opened 2022-06-10
naywatch: fix naywatch

PolynomialDivision:
We need to give the interface names (mgmt) instead of the device names
(lan0.42).


PolynomialDivision:
You can now check if naywatch is active:
https://github.com/openwrt/routing/commit/09d5ceb9230b907029a3e09858ef181ff85e3913

PR #236, opened 2022-06-16
linsys-e8450: fix building images

PolynomialDivision:
OpenWrt moved wolfssl from packages to main repository. However, the
package feed is now invalid. Go to openwrt-snapshot again.


PolynomialDivision:
I will merge. The imagebuilder is broken in 22.03-rc1 and this will not change in future.

PR #239, opened 2022-06-17
segen: more accurate geo position

Noki:
The current position is inaccurate and hopglass shows the position marker on the street. This fixes the position so it points to the center of the tower of the church.


PolynomialDivision:
lgtm

Akira25 (approved):

PR #250, opened 2022-06-22
add helper script to generate images

Noki:
This is a small helper script that generates a menu based on the locations and generates images based on number selection.


PolynomialDivision:
applied to master, thanks!
https://github.com/freifunk-berlin/bbb-configs/commit/d8dba7b80ec3c39eb016c3359e8434be6ee1c743

PR #251, opened 2022-06-22
add yamllint as a standalone linter with its own config

Noki:
ansible-lint uses yamllint to lint yaml files but does not allow a custom config for yamllint e.g. to configure or ignore line length. this commit fixes this by adding yamllint as a standalone linter with its own config


PolynomialDivision:
lgtm. thanks!

PR #252, opened 2022-06-22
fix most yamllint issues

Noki:


Akira25 (approved):
Looks good to me. There are some yes-vs.-true-issues left, but we could address tem in a separate PR.

PR #266, opened 2022-06-29
conditional_packages: add extended_privacy flag

PolynomialDivision:


robertfoss:
Are the MACs transmitted outside of the router? Or are they just used for counter, and then the count is transmitted?

Noki:
Personally I liked the previous PR better just because I think we should have it the other way around and make the station node exporters an optional opt-in. So I propose:

* remove the packages from the default list
* add the packages if "station_statistics" or something similar is set

We should call it something like: "conditional_packages: make packages with privacy concerns an opt-in"

PR #268, opened 2022-06-30
fix shellcheck issues

Noki:


PolynomialDivision:
runtested. works perfectly fine.

PR #274, opened 2022-07-05
Add suedblock

ideesnoires:
build, deployed and tested. this is adding suedblock into the repo. had still an old image version kathleen and no available login credentials anymore


PolynomialDivision:
Thanks. Applied in 6ee215a0e79292a714f187de844aec551b726c73

PR #277, opened 2022-07-07
extended DEVELOPERS.md with tunnel config

ideesnoires:
just added an example block for wireguard tunnels

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -130,6 +130,23 @@ The VLAN ID (vid) usually follow this numbering convention.
 42  MGMT
 ```
 
+##### uplink via tunnel
+
+if for some reason you are in need of an uplink via a "normal" internet connection, a wireguard
+tunnel can be an easy and safe solution. in that case add another vlan to the networks.yml
+
+```yml
+  - vid: 13
+    untagged: true            # option to don't tag this vlan - useful if the corerouter is plugged into a normal home router
+    name: uplink
+    role: ext
+    tunnel_wan_ip: 192.168.1.2/24   # put here the address and subnet of the corerouter inside the uplink network
+    tunnel_wan_gw: 192.168.1.1      # gateway of the uplink network
+    tunnel_connections: 2           # default value, number of different tunnels to create
+    tunnel_timeout: 600             # timeout in seconds after this the tunnel is destroyed and attempted to be rebuild
+    tunnel_mesh_prefix_ipv4: 10.31.142.120/30   # ip subnet to pick addresses for the endpoints of the tunnels
+```
+
 #### ssh-key.yml
 
 By default the ssh-keys within `all/ssh-keys.yml` will be installed on all hosts. To add additional ssh keys a ssh-key.yml file can be created with additional keys in the following format:


DEVELOPER.md
```diff
@@ -114,6 +114,17 @@ networks:
     assignments:              # assign static addresses to devices
       magda-core: 1
 
+# not part of magda conf, only to illustrate tunnel usage
```
Noki, line 117: please put the example in a seperate code block

DEVELOPER.md
```diff
@@ -114,6 +114,17 @@ networks:
...
+  - vid: 13
+    untagged: true            # don't tag this vlan - useful if the corerouter is plugged into a normal home router
+    name: uplink
+    role: ext
+    tunnel_wan_ip: 192.168.1.2/24   # put here the address of the corerouter inside the uplink network
+    tunnel_wan_gw: 192.168.1.1      # gateway of the uplink network
+    tunnel_connections: 2           # default value, number of different tunnels to create
+    tunnel_timeout: 600             # timeout in seconds after this the tunnel is destroyed and attempted to be rebuild
+    tunnel_mesh_prefix_ipv4: 10.31.142.120/29   # ip address space for both sides of the tunnel
```
Noki, line 126: This should be explained in more detail. What is the size requirement for this ip-block? Is it tunnel_connections * 2? In addition it should also clearly state that this is a subnet derived from the main ip-block for this node.

DEVELOPER.md
```diff
@@ -114,6 +114,17 @@ networks:
     assignments:              # assign static addresses to devices
       magda-core: 1
 
+# not part of magda conf, only to illustrate tunnel usage
+  - vid: 13
+    untagged: true            # don't tag this vlan - useful if the corerouter is plugged into a normal home router
```
Noki, line 119: can this also be a tagged vlan? can I limit this vlan to a single port?

DEVELOPER.md
```diff
@@ -114,6 +114,17 @@ networks:
...
+    name: uplink
+    role: ext
+    tunnel_wan_ip: 192.168.1.2/24   # put here the address of the corerouter inside the uplink network
```
Noki, line 122: this is a network, not only an ip. I suggest to change this to "put the addess of the core router inside the uplink network here together with the netmask of the uplink network.

Noki (changes requested):

DEVELOPER.md
ideesnoires: i put there because it's part of the networks.yml so how can i put it into an extra block?

DEVELOPER.md
ideesnoires: was technically wrong, now fixed. this is only the ip of the corerouter end of the tunnel

DEVELOPER.md
ideesnoires: explained a bit more. can also be tagged. afaik not possible to assign to a specific port.

DEVELOPER.md
ideesnoires: i mean should i put it into an extra section like: uplink via tunnel -> your networks.yml is gonna look like this?

DEVELOPER.md
```diff
@@ -130,6 +130,23 @@ The VLAN ID (vid) usually follow this numbering convention.
...
+```yml
+  - vid: 13
+    untagged: true            # option to don't tag this vlan - useful if the corerouter is plugged into a normal home router
+    name: uplink
+    role: ext
+    tunnel_wan_ip: 192.168.1.2/24   # put here the address and subnet of the corerouter inside the uplink network
+    tunnel_wan_gw: 192.168.1.1      # gateway of the uplink network
+    tunnel_connections: 2           # default value, number of different tunnels to create
+    tunnel_timeout: 600             # timeout in seconds after this the tunnel is destroyed and attempted to be rebuild
+    tunnel_mesh_prefix_ipv4: 10.31.142.120/32   # ip address for this side of the tunnel
```
PolynomialDivision, line 147: This config will not work. You have a "pool" of 1 address because of /32 but you want 2 tunnel_connections.

Noki (approved):

PolynomialDivision:
Please squash the commits together

PR #283, opened 2022-07-12
Renumber IPv4+6 address on philmel

booo:
This completes the move to the new ansible configuration for phimel.

Please wait with merge till I was able to test the setup on a local router.


Akira25:
@booo did you set it up already?

booo:
Yes. As far as I can tell the router configuration works and we can merge this.

PR #294, opened 2022-07-24
huette15: new corerouter, separate ap, new addresses, vdsl uplink

pktpls:


group_vars/location_huette15/networks.yml
```diff
@@ -1,39 +1,69 @@
...
+
+  - vid: 50
+    role: ext
+    name: easybell
+    untagged: true
+    tunnel_wan_ip: 192.168.178.2/24
+    tunnel_wan_gw: 192.168.178.1
+    tunnel_connections: 2
+    tunnel_timeout: 600
+    tunnel_mesh_prefix_ipv4: 10.31.82.192/28
```
Noki, line 63: 2 tunnel connections will use 2 addresses of that block, but you already use the first 2 for vid 20 and vid 21. I think you need to split the net.

pktpls: It checks which addresses are already in use and picks ones which are free :-)

```
13: wg_51820: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1412 qdisc noqueue state UNKNOWN group default qlen 1000
    link/none 
    inet 10.31.82.192/32 scope global wg_51820
       valid_lft forever preferred_lft forever
    inet6 fe80::2/64 scope link 
       valid_lft forever preferred_lft forever
14: wg_51821: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1412 qdisc noqueue state UNKNOWN group default qlen 1000
    link/none 
    inet 10.31.82.195/32 scope global wg_51821
       valid_lft forever preferred_lft forever
    inet6 fe80::2/64 scope link 
       valid_lft forever preferred_lft forever

```

(Although it probably shouldn't pick 192 in this case, but that's another matter.

PR #295, opened 2022-07-25
k11: added and made k12 connection, fixed model definitions

Noki:
- VLANs are in the 1xx range so k11 can get connected to k12 with a cable without causing problems and then be merged together.
- Models needed packages changed to be able to mesh via 802.11s


Akira25 (approved):
Looks good to me on the first sight.

PR #298, opened 2022-07-31
corerouter: add lldpd

PolynomialDivision:
We need to get independent of the olsrd name protocol. Add lldpd as
standatarized protocol to get information about neighbors. To test it
just do

  lldpcli show neighbors

We can later use the information to feed owm with it.

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/config/lldpd.j2
+++ b/roles/cfg_openwrt/templates/corerouter/config/lldpd.j2
@@ -1,0 +1,12 @@
+#jinja2: trim_blocks: "true", lstrip_blocks: "true"
+config lldpd config
+	option enable_cdp 0
+	option enable_fdp 0
+	option enable_sonmp 0
+	option enable_edp 0
+	option lldp_class 4
+	list interface 'loopback'
+{% for network in networks | selectattr('role', 'equalto', 'mesh') %}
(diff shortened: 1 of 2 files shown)


roles/cfg_openwrt/templates/corerouter/config/lldpd.j2
```diff
@@ -0,0 +1,13 @@
+#jinja2: trim_blocks: "true", lstrip_blocks: "true"
+config lldpd config
+	option enable_cdp 0
+	option enable_fdp 0
+	option enable_sonmp 0
+	option enable_edp 0
+	option lldp_class 4
+	option lldp_location "2:FR:6:Commercial Rd:3:Roseville:19:4"
```
Noki, line 8: This should be properly filled with the location of the device.

PolynomialDivision: I think we can just delete this line.

Noki (approved):
I tested this and the data and output-formats look really good. I think it is a great idea to roll this out over all bbb-configs corerouter nodes so we can play around with the data. My only concern is that the package is quite big (131 KB).

Akira25:
Before applying this, we should definitely discuss this in a meeting!

spolack:
I'd also opt for not merging this at the moment. But we definitely should get documented the findings and pro/con arguments for using this over http endpoint.

PR #301, opened 2022-08-09
Add the usage of a venv to the documentation

rtznprmpftl:
I added a quick info in regards to a virtualenv, since this should be the default when using python.

This way you don't collect dependencies from random projects in your homedir

### DIFF
--- a/.config/ansible-lint.yml
+++ b/.config/ansible-lint.yml
@@ -1,7 +1,8 @@
 ---
 exclude_paths:
-  - .github/
-  - .git
+  - ../.github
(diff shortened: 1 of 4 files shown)


rtznprmpftl:
Added the venv to the ansible lint config & fixed the old excluded_paths

# exclude_paths included in this file are parsed relative to this file's location
# and not relative to the CWD of execution. CLI arguments passed to the --exclude
# option will be parsed relative to the CWD of execution.

Noki:
looks good to me. tested and works as described.

PR #303, opened 2022-08-10
treewide: use cdn.openwrt for downloads

PolynomialDivision:
This will reduce the load on openwrt servers.


pktpls:
Sounds good! :+1:

For reference, more information about the cdn mirror in openwrt/openwrt@c737a9ee6a9c47b6e553ac81bf293b1161e59799

PR #304, opened 2022-08-11
Ak36 vm

pmelange:
Reconfigure ak36-gw as a virtual machine


pktpls:
Good to see ak36-gw back up!

And also good to see the x86-64 target used, now we have example code for the VLAN :-)

Akira25 (approved):
Looks good to me.

PR #307, opened 2022-08-15
Rixbox init

pktpls:
Rixbox Espresso & Food @ Karl-Marx-Straße / Alfred-Scholz-Platz - very nice MABB site from 2014 that got new gear.

This soft-depends on blocktrron/poemgr#5 - it builds correctly without this patch, but fails to power up all devices. We can merge anyway - we give priority to the Rixbox-RHNK uplink and leave the APs disabled. Then after first boot, we install the patched poemgr package and enable the APs.

The wiki has been updated to reflect the new setup: https://wiki.freifunk.net/Berlin:Standorte:Rixbox

### DIFF
--- a/group_vars/model_ubnt_usw_flex.yml
+++ b/group_vars/model_ubnt_usw_flex.yml
@@ -21,6 +21,8 @@ model__packages__to_merge:
   - "-prometheus-node-exporter-lua-snmp6"
   - "-prometheus-node-exporter-lua-wifi"
   - "-prometheus-node-exporter-lua-wifi_stations"
+  # TODO reinstate this package once it's fixed in 1.3.0-snapshot feed
+  - "-luci-app-falter-owm-ant"
 
 compat_version: 1.1
 dsa_ports:
--- a/group_vars/model_ubnt_usw_flex.yml
+++ b/group_vars/model_ubnt_usw_flex.yml
@@ -1,0 +1,31 @@
+---
+
+# Factory flash instructions:
+# https://github.com/openwrt/openwrt/commit/a9839697896c4fdf8c44a06bbce466ce52493069
+
+openwrt_version: 22.03-SNAPSHOT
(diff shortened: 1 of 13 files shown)


group_vars/model_ubnt_usw_flex.yml
```diff
@@ -0,0 +1,44 @@
...
+# Factory flash instructions:
+# https://github.com/openwrt/openwrt/commit/a9839697896c4fdf8c44a06bbce466ce52493069
+
+openwrt_version: snapshot
+target: ramips/mt7621
+
+# Our partition is only 7360k, not 8m.
+# We actually need to be below 7200k as the next squashfs padding step is 7400k.
+low_flash: true
+model__packages__to_merge:
```
PolynomialDivision, line 12: why do you delete all that packages?

group_vars/model_ubnt_usw_flex.yml
```diff
@@ -0,0 +1,33 @@
...
+feed_version: 1.3.0-snapshot
+target: ramips/mt7621
+
+# Our partition is only 7360k, not 8m.
+# We actually need to be below 7200k as the next squashfs padding step is 7400k.
+low_flash: true
+model__packages__to_merge:
+  - "-tcpdump-mini"
+  - "-curl"
+  - "-prometheus-node-exporter-lua"
```
PolynomialDivision, line 16: Can't you remove something else? Monitoring plugins are essential. :/

pktpls: They depend on `curl` though which is 200 KB (!) :smiling_face_with_tear: I'd rather remove these, than use `low_flash` and lose OPKG, because then I can't install the patched poemgr package.

We also still have Collectd.

PolynomialDivision: okay :)

PR #309, opened 2022-08-20
usw-flex: we require opkg and don't need low_flash

pktpls:
Somehow Github didn't pick up this last change I pushed last night...

We don't need `low_flash` since removing `curl` saves enough disk space. We also require `opkg` which would be removed by `low_flash`.


Noki:
looks good to me

PR #310, opened 2022-09-01
dragonkiez-buero

pmelange:

### DIFF
--- a/group_vars/location_dragonkiez_buero/networks.yml
+++ b/group_vars/location_dragonkiez_buero/networks.yml
@@ -1,0 +1,29 @@
+---
+ipv6_prefix: "2001:bf7:830:b300::/56"
+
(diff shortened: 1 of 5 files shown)


group_vars/location_dragonkiez_buero/networks.yml
```diff
@@ -0,0 +1,29 @@
...
+    assignments:
+      dragonkiez-buero: 1
+      dragonkiez-buero-rhnk: 2
+
+  - vid: 40
+    role: dhcp
+    untagged: true
+    inbound_filtering: false
+    enforce_client_isolation: false
+    prefix: 10.31.177.161/27
```
Noki, line 26: This does not match the DHCP network within the comments. I know it's the same network, but is it intentional?

PR #314, opened 2022-09-05
bht: correct Perrys Mail

Akira25:
This one was missed for quite a time. Please don't miss the
'i' at the beginning of the address! Using vim is no excuse
for that.


Noki:
looks good to me

PR #320, opened 2022-09-10
Fix contacts

Akira25:
### contacts flag:
Setting the contacts flag will integrate any type of contact option in the
generated config. It must be a list and enables us to give several options
beside email addresses. One may also give a link to a matrix room, icq number
or whatever.

Example:
```yml
---
contact_name: 'Mr. Good-Code'
contact_nickname: 'gc-guy'
contacts:
  - 'some_address@example.org'
  - '01234 567890'
  - 'info@berlin.berlin'
  - 'https://matrix-raum.to'
...
```

### community-flag:
By setting 'community: true' in the locations general.yml, we can mark a site
as maintained by the bbb-configs administrator-team. This will fill the mail-field
in the configuration automatically with an email-address in the format
info+SITE_NAME@berlin.freifunk.net. This may be overridden by defining the
contacts flag

## Note:

This PR depends on https://github.com/freifunk-berlin/falter-packages/pull/330

### DIFF
--- a/roles/cfg_openwrt/templates/common/config/freifunk.j2
+++ b/roles/cfg_openwrt/templates/common/config/freifunk.j2
@@ -2,8 +2,18 @@ config public 'contact'
 {% if contact_name is defined %}
 	option name '{{ contact_name }}'
 {% endif %}
+{% if nickname is defined %}
 	option nickname '{{ contact_nickname }}'
-	option mail '{{ contact_email }}'
+{% endif %}
+{% if contact_email is defined %}
+	{{ (contact_email_depr) | mandatory('The option "contact_email" in location/general.yml is deprecated. Please rename that flag to "contacts" and make sure to transform it into to a list.') }}
+{% elif (community is true) and (contacts is undefined)%}
+	option mail 'info+{{ inventory_hostname }}@berlin.freifunk.net'
+{% else %}
+{% for contact in (contacts) | mandatory('Please add at least one contact information via the contacts-flag in location/general.yml. Fellow Freifunkas need to be able to contact you in case they want to mesh with you!') %}
+	list contact '{{ contact }}'
+{% endfor %}
+{% endif %}
 	option location '{{ location_nice }}'
 
 config public 'community'
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -53,9 +53,10 @@ This file might be fairly self-explanatory. Please mind that a contact is mandat
 ---
 contact_name: 'Petrosilius Quaccus'
 contact_nickname: 'Petro'
-contact_email: 'quaccus@example.com'
-# or:
-contact_email: 'https://config.berlin.freifunk.net/contact/446x/ImZmZnctZV0LTA0Ig.FFbQ8w._ZCA4hFY3zR8MdDVNrv3okqwPU'
+# contacts must be a list. Even if only one contact is given
+contacts:
+  - 'quaccus@example.com'
+  - 'https://config.berlin.freifunk.net/contact/446x/ImZmZnctZV0LTA0Ig.FFbQ8w._ZCA4hFY3zR8MdDVNrv3okqwPU'
 ```
 
 #### owm.yml
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -2,10 +2,6 @@
 zonename: 'Europe/Berlin'
 timezone: 'CET-1CEST,M3.5.0,M10.5.0/3'
 
-contact_nickname: 'Freifunk Berlin Maintenance Team'
-contact_email: 'info@berlin.freifunk.net'
-
-
 dns_servers:
   - 46.182.19.48
   - 80.67.169.40
--- a/play.yml
+++ b/play.yml
@@ -2,5 +2,6 @@
 - hosts: all
   connection: local
   gather_facts: false
+  any_errors_fatal: true
   roles:
     - {role: cfg_openwrt}


Akira25:
For the failing CI: The Linting-Errors look like they belong to another file. At least I cannot see what the error prompt is trying to tell me.

The test-build is failing by intention: The tested sites don't comply with the introduced scheme. Adjusting them to it should fix it easily.

roles/cfg_openwrt/templates/common/config/freifunk.j2
```diff
@@ -2,8 +2,18 @@ config public 'contact'
 {% if contact_name is defined %}
 	option name '{{ contact_name }}'
 {% endif %}
+{% if nickname is defined %}
 	option nickname '{{ contact_nickname }}'
-	option mail '{{ contact_email }}'
+{% endif %}
+{% if contact_email is defined %}
```
Noki, line 8: This should be implemented differently so it is backwards compatible:
If contact_email is set it should simply be used as an entry in contacts. If the site has the community flag and contacts is empty the default should be used. If it is absent AND the community flag is false it should fail.

Akira25: > If contact_email is set it should simply be used as an entry in contacts.

We could do it that way. But in my opinion, allowing (legacy) email-contacts-field might introduce unnecessary complexity. My intention was to encourage each site maintainer to migrate their code to the new standard by their self, once they build the site next time. That way, we can have migrated the whole repo in a year or so.

> If the site has the community flag and contacts is empty the default should be used.

In the mentioned [wednesday meeting](https://lists.berlin.freifunk.net/pipermail/berlin/2022-September/053153.html) we agreed on plus-addressing. It has some advantages, like we can sort the incoming emails better. All Emails addresses to info+SOMETHING@berlin will go to the regular info-address. So info+SITE@berlin should be the new standard. There's no change in the current receiving infrastructure.


> If it [ contacts-field ] is absent AND the community flag is false it should fail.

That's already happening, isn't it? Do you think, we should require the community flag to be set explicitly to false, instead of leaving it undefined?

Noki: > > If contact_email is set it should simply be used as an entry in contacts.
> 
> We could do it that way. But in my opinion, allowing (legacy) email-contacts-field might introduce unnecessary complexity. My intention was to encourage each site maintainer to migrate their code to the new standard by their self, once they build the site next time. That way, we can have migrated the whole repo in a year or so. Breaking backward compability was by intention here.

Ok, got the idea. But in that case please migrate the contact_email fields from all existing locations to the contacts list together with the pull request and add the community flag for the community locations. From what I can see (`grep -r "contact_email" .` within `group_vars`) most configurations are just fine and forcing all those people to migrate instead of just migrating for them is bad style.

> 
> > If the site has the community flag and contacts is empty the default should be used.
> 
> In the mentioned [wednesday meeting](https://lists.berlin.freifunk.net/pipermail/berlin/2022-September/053153.html) we agreed on plus-addressing. It has some advantages, like we can sort the incoming emails better. All Emails addresses to info+SOMETHING@berlin will go to the regular info-address. So info+SITE@berlin should be the new standard. There's no change in the current receiving infrastructure.

With this I ment the proposed new default (info+SITE@berlin).

> 
> > If it [ contacts-field ] is absent AND the community flag is false it should fail.
> 
> That's already happening, isn't it? Do you think, we should require the community flag to be set explicitly to false, instead of leaving it undefined?

The idea here was that a site must be a community site (contact set or contact mail generated as proposed) or if not setting a contact field is mandatory.

In general I think having the community flag is a really good idea as it would allow us to differentiate sites that can be mass flashed by the community vs. sites that are operated by a single person and can not or should not be mass flashed. I think the community flag should be set to true only for the community sites and should otherwise be absent.

Noki (approved):
Looks good to me. Since this is a breaking change and I did not attend the meeting where this was discussed, I would love to see another approval.

roles/cfg_openwrt/templates/common/config/freifunk.j2
```diff
@@ -2,8 +2,18 @@ config public 'contact'
...
+	{{ (contact_email_depr) | mandatory('The option "contact_email" in location/general.yml is deprecated. Please rename that flag to "contacts" and make sure to transform it into to a list.') }}
+{% elif (community is true) and (contacts is undefined)%}
+	option mail 'info+{{ inventory_hostname }}@berlin.freifunk.net'
+{% else %}
+{% for contact in (contacts) | mandatory('Please add at least one contact information via the contacts-flag in location/general.yml. Fellow Freifunkas need to be able to contact you in case they want to mesh with you!') %}
```
spolack, line 13: Why do you put references to variables in () ?

roles/cfg_openwrt/templates/common/config/freifunk.j2
```diff
@@ -2,8 +2,18 @@ config public 'contact'
 {% if contact_name is defined %}
 	option name '{{ contact_name }}'
 {% endif %}
+{% if nickname is defined %}
```
spolack, line 5: shouldnt be contact_nickname?

group_vars/location_chris/general.yml
```diff
@@ -0,0 +1,3 @@
+---
+contact_nickname: 'FETF – Freifunk Engenieering Task Force'
```
spolack, line 2: I'd prefer to also set that contact nickname by default, if thats a community site.

roles/cfg_openwrt/templates/common/config/freifunk.j2
Akira25: I think in that special case the reason is, that the mandatory directive should work on the `contacts`-variable only. I think I've seen that somewhere while googling. Not sure, if it will work the same without the brackets.

roles/cfg_openwrt/templates/common/config/freifunk.j2
Akira25: your right, I'll change that.

Akira25:
@spolack I've made the proposed changes. Is it okay for you to merge this?

spolack:
LGTM, but as discussed via PN please remove the (). Feel free to merge then. Thanks for your work!

PR #325, opened 2022-09-24
add support for wlan band replacing hwmode

Noki:
As described in the OpenWRT documentation (https://openwrt.org/docs/guide-user/network/wifi/basic) hwmode was replaced by band in 21.02.2. This PR introduces support for using band instead of hwmode and updates the documentation accordingly. For easy transitioning if present band will be used instead of hwmode, but hwmode will still work.


Akira25:
Looks good to me.

PR #326, opened 2022-09-24
Add htmode prefix

### LINKED ISSUE #261: Configure wifi mode correctly
We should use the highest supported mode. N,AC,AX.

Noki:
This pull request adds support to specify an htmode_prefix that allows to specify the highes possible htmode for the wlan device. If accepted this should close #261.

What's missing is adding the setting for all models.


spolack (commented):
LGTM

PR #328, opened 2022-09-28
gateways: replace fw3 with fw4, translate all rules to nftables; treewide: switch to 22.03

spolack:
This is running on saarbruecker-gw now.

Interesting commands to troubleshoot:
```
fw4 print
nft list ruleset

```


roles/cfg_openwrt/templates/uplink_gateway/config/firewall.j2
```diff
@@ -5,6 +5,7 @@ config defaults
 	option output		ACCEPT
 	option forward		REJECT
 
+
```
PolynomialDivision, line 8: why?

spolack: I'll remove that blank new line

spolack:
I already tested that change on magda-core which is ipq40xx and magda-ap1.

spolack:
Let's wait until https://github.com/openwrt/openwrt/pull/10847 is merged. That will hopefully allow us to run the 8M/32M Devices with current OpenWRT

PolynomialDivision:
I would propose to merge and just keep the ubiquiti xm devices to 21.02. No idea when and if that is going to be merged.

spolack:
Well, then lets lets use snapshot for them.

PolynomialDivision (approved):

PR #329, opened 2022-09-28
wireless devices: replace hwmode with band, set htmode

spolack:
This is a follow up PR for #325 and #326


Noki (approved):
looks good to me

PR #334, opened 2022-10-01
documentation: add metrics for tunnelmanager & ssh-keys replacement

Noki:


Akira25 (approved):
Looks good to me.

Akira25 (approved):
Looks good to me again.

Maybe we shouldn't name that file `developer.md`, because it contains information which is interesting for users too. But thats another issue.

PR #340, opened 2022-10-03
Make APs use NTP server of the core router

### LINKED ISSUE #4: Implement Core Router to act NTP Server and make aps to use it.
We should enable ntp server per default on the core routers and find a clean way how to specify that the core router shall be used as ntp server.

maybe similar to how its done for DNS and gateway in the network definition in the location group vars.

Noki:
This pull request should address the remaining stuff to close #4.

To actually start using the NTP server on the core router the corresponding networks files have to be changed later on.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -103,6 +103,7 @@ networks:
     prefix: 10.31.83.112/28
     gateway: 1
     dns: 1
+    ntp: 1                    # used to tell accesspoints to use the ntp server of the core router
     ipv6_subprefix: 1
     assignments:              # assign static(!) addresses from mngt-network to individual devices/interfaces.
       magda-core: 1           # core router gets 1st address. In result it will be reachable at 10.31.83.113
--- a/roles/cfg_openwrt/templates/common/config/system.j2
+++ b/roles/cfg_openwrt/templates/common/config/system.j2
@@ -18,9 +18,16 @@ config system
 config timeserver 'ntp'
         option enabled '1'
         option enable_server '{{ (role == 'corerouter' or role  == 'uplink_gateway') | int }}'
+{% if networks is defined %}
+{% set mgmtnetwork = networks | selectattr('role', 'equalto', 'mgmt') | default([], true) | list | first %}
+{% endif %}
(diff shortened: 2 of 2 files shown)


pktpls:
Today I learned that there's a multicast group just dedicated to NTP servers. If we can get the corerouter to join that group (ff02::101), then getting the correct server address is super easy and we don't need to use VLAN 42 for it.

https://www.iana.org/assignments/ipv6-multicast-addresses/ipv6-multicast-addresses.xhtml

DEVELOPER.md
```diff
@@ -103,6 +103,7 @@ networks:
     prefix: 10.31.83.112/28
     gateway: 1
     dns: 1
+    ntp: 1                    # used to tell accesspoints to use the ntp server of the core router
```
pktpls, line 106: We can just make this the default, no?

pktpls:
LGTM, go ahead and merge.

I tried out the ff02::101 thing - but ntpd in OpenWrt is so minimal that it doesn't do any IPv6 at all...

DEVELOPER.md
Noki: I just mimicked what we have for dns and gateway and I would prefer to have it set and visible in the configuration. On some locations you might even want to change this to another time server e.g. an access point with build in GPS.

roles/cfg_openwrt/templates/common/config/system.j2
```diff
@@ -18,9 +18,16 @@ config system
 config timeserver 'ntp'
         option enabled '1'
         option enable_server '{{ (role == 'corerouter' or role  == 'uplink_gateway') | int }}'
+{% if networks is defined %}
+{% set mgmtnetwork = networks | selectattr('role', 'equalto', 'mgmt') | default([], true) | list | first %}
```
spolack, line 22: IMO not very elegant. I'll try to make another proposal these days. Reset LGTM

Akira25 (approved):
Looks good to me. As this issue waits for quite some time now, and there weem to be no functional blocking issues, I'd merge this. Please feel free to revert this change or make an additional PR, if you disagree on that assesment.

PR #350, opened 2022-10-12
kirchhof: init

pktpls:


group_vars/all/wireless_profiles.yml
```diff
@@ -312,6 +312,76 @@ wireless_profiles:
...
+        network: dhcp
+        radio: [11a_standard, 11g_standard]
+        ifname_hint: ff
+      - mode: ap
+        ssid: berlin.freifunk.net
+        encryption: owe
+        network: dhcp
+        radio: [11a_standard, 11g_standard]
+        ifname_hint: ffowe
+        ieee80211w: 1
```
pktpls, line 377: We'll try out OWE with "berlin.freifunk.net" SSID for a while. After that we want to try Transition Mode.

spolack: Are you aware that iphones have to use "unsafe" wifi then?

pktpls:
We're going to revert this to the traditional SSID naming. Still a few annoying bugs with WPA3 OWE and Intel cards...

PR #351, opened 2022-10-12
k11: Change prefixes

Akira25:
Due to a collision with hand-crafted Museumsinsel, we change the prefixes to be unique again.

### DIFF
--- a/group_vars/location_k11/networks.yml
+++ b/group_vars/location_k11/networks.yml
@@ -2,11 +2,12 @@
 ipv6_prefix: '2001:bf7:760:100::/56'
 
 # got following prefixes:
-# Router:  10.31.17.0/24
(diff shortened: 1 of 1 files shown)


group_vars/location_k11/networks.yml
```diff
@@ -2,19 +2,19 @@
 ipv6_prefix: '2001:bf7:760:100::/56'
```
Noki, line 2: pretty sure this has to change as well... IPv4 and IPv6 were registered together and should be again...

Akira25: Jep, forgot to include them in the request.

Akira25:
@Noki Finally got the time for the database-crawling stuff. If you are curious on the ipv6-network: Apparently we really got the same network again:

grafik

PR #354, opened 2022-10-17
k12: disable uplink due to routing problems

Noki:


PolynomialDivision:
Whats happening?

Noki:
@PolynomialDivision IPv6 traffic is going through the tunnel that should act only a backup uplink. Read the matrix discussion about this: https://matrix.to/#/!XqRjbdZgEEQXZclmsI:matrix.org/$9PsBDhEtRFX17_vcK95Jbi5ZBvd42JywGBsstSPAhTA?via=matrix.org&via=systemli.org&via=matrix.riotcat.org
simzon pointed out the problem an what needs to be done, however I don't really feel up for the task.

PR #357, opened 2022-10-23
gub37: Add gub37-hof-w radio

robertfoss:


spolack:
LGTM.

Two thoughts:
* Do you really want to use 80 MHz channels? The backbone links are 40MHz wide, you hardly wont get any more throughput. Depending on how far your clients are also narrower channel width make sense, because it should improve the RX sensitivity, which could lead to higher modulation levels.
* I would appreciate if you could test the `freifunk_default_owe_transition` profile. For me its working just fine and i'd need more feedback before making it the new default.

spolack (approved):

robertfoss:
@spolack The 80Mhz test links are are all kind of high quality & very short distance. For both hof-n & hof-w, so the additional bandwidth does not seem to have much of a downside.

@spolack  About `freifunk_default_owe_transition` sure, I don't mind testing it. This location is not very public, just so you know.

PR #358, opened 2022-10-24
Add DAWN to support band-steering for large nodes

robertfoss:
Larger nodes often have multiple radios serving the same SSIDs. In order to improve end-user experiences, enable band-steering using Dawn for `corerouter` and `ap` roles.

This branch has been deployed successfully at the gub37 & gruni73 nodes.

### DIFF
--- a/host_vars/gub37-core/base.yml
+++ b/host_vars/gub37-core/base.yml
@@ -5,3 +5,4 @@ role: corerouter
 model: "avm_fritzbox-7530"
 
 wireless_profile: freifunk_default
+wifi_roaming: true
--- a/roles/cfg_openwrt/tasks/conditional_packages.yml
+++ b/roles/cfg_openwrt/tasks/conditional_packages.yml
@@ -12,6 +12,19 @@
     - airos_dfs_reset|default([])|length > 0
     - role == 'corerouter'
 
+- name: "Add Dawn + Luci app for corerouters"
+  set_fact:
+    packages: "{{ packages + [item] }}"
+  when:
+    - role == 'corerouter' or role == 'ap'
+    - not (low_mem | default(false))
+    - not (low_flash | default(false))
+    - wifi_roaming | default(false)
+  loop:
+    - dawn
+    - luci-app-dawn
+    - umdns
+
 - name: "Add tunnelmanager if configured"
   set_fact:
     packages: "{{ packages + ['falter-berlin-tunnelmanager'] }}"
--- a/host_vars/gub37-hof-s/base.yml
+++ b/host_vars/gub37-hof-s/base.yml
@@ -3,3 +3,5 @@
 location: gub37
 role: ap
 model: "mikrotik_sxtsq-5-ac"
+
+wifi_roaming: true
--- a/roles/cfg_openwrt/templates/ap/config/dawn.j2
+++ b/roles/cfg_openwrt/templates/ap/config/dawn.j2
@@ -1,0 +1,83 @@
+{% set network_mgmt = networks | selectattr('role', 'equalto', 'mgmt') | first %}
+{% set mgmt_broadcast = network_mgmt['prefix'] | ansible.utils.ipaddr('broadcast') %}
+{% if wifi_roaming | default(false) %}
+config local
+	option loglevel '0'
+
+config network
+	option broadcast_ip '{{ mgmt_broadcast }}'
+	option broadcast_port '1025'
+	option tcp_port '1026'
+	option network_option '2'
+	option shared_key 'dawn@{{ location_nice|default(location) }}'
+	option iv 'thisisaniv!'
+	option use_symm_enc '0'
+	option collision_domain '-1'
+	option bandwidth '-1'
+
+config hostapd
+	option hostapd_dir '/var/run/hostapd'
+
+config times
+	option con_timeout '60'
+	option update_client '10'
+	option remove_client '15'
+	option remove_probe '30'
+	option remove_ap '460'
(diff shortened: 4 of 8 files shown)


robertfoss:
As of today this is deployed and functional at gruni73.

spolack:
Looking okayish at a first glance. I have to dive deeper in order to really understand what the config does and whether more abstraction would be beneficial. I'll try do review within the next weeks.

spolack:
Please implement at least some flag which controls, whether DAWN should be enabled for a location or not.

robertfoss:
Added the `bandsteering` flag, which installs dawn + dependencies.

However! If this flag is disabled, the `/etc/config/dawn` & `/etc/config/umdns` config files are still generated from the ansible/jinja template files. As far as I can tell there is no way to ignore specific template files if certain flags are set.

host_vars/gub37-core/base.yml
```diff
@@ -5,3 +5,4 @@ role: corerouter
 model: "avm_fritzbox-7530"
 
 wireless_profile: freifunk_default
+bandsteering: true
```
PolynomialDivision, line 8: I would not call it bandsteering. Depending on DAWN configuration it does several things, for example, client load balancing.

robertfoss: Calling it dawn felt to specific. Do you have a suggestion?

roles/cfg_openwrt/tasks/conditional_packages.yml
```diff
@@ -12,6 +12,19 @@
     - airos_dfs_reset|default([])|length > 0
     - role == 'corerouter'
 
+- name: "Add Dawn + Luci app for corerouters"
+  set_fact:
+    packages: "{{ packages + [item] }}"
+  when:
+    - role == 'corerouter' or role == 'ap'
```
robertfoss, line 19: Doesn't this line basically make sure that all routers + access points in a site will have dawn enabled (if it's been set to enabled)?

PolynomialDivision: I thought only, when bandsteering is enabled in the host? (see the 3 lines to the bottom).

robertfoss: The Corerouter  & the APs?

spolack:
> However! If this flag is disabled, the /etc/config/dawn & /etc/config/umdns config files are still generated from the ansible/jinja template files. As far as I can tell there is no way to ignore specific template files if certain flags are set.

Yes, please generate empty config file instead of actually templating some unused config.

spolack:
In addition is see the problem that VLAN 42 is sometimes bridged across multiple locations, we should ensure the dawns dont talk to each other. Any idea how to solve that?

spolack:
We should also use the same config for APs / Core Routers -> Use Symlink

Thats a local changeset further tweaking the settings
```
+++ b/roles/cfg_openwrt/templates/corerouter/config/dawn.j2
@@ -38,13 +38,13 @@ config metric 'global'
        option eval_auth_req '0'
        option eval_assoc_req '0'
        option kicking '3'
-       option kicking_threshold '20'
+       option kicking_threshold '10' # Make it more likely to prefer the Better AP
        option deny_auth_reason '1'
        option deny_assoc_reason '17'
        option min_number_to_kick '3'
        option chan_util_avg_period '3'
        option set_hostapd_nr '0'
-       option duration '0'
+       option duration '5' # This is required, otherwise it's not kicking, suggested by nick
        option rrm_mode 'pat'
```

robertfoss:
@PolynomialDivision I switched variable name from `bandsteering` to `wifi_roaming`.

@spolack
 - Fixed vlan42 issue by setting the dawn passphrase to the location name
 - Fixed template generating unused config files, now the file is empty
 - Tested using symlinks, but ansible does not seem to deal with that gracefully, log below
 
 
 ```
 An exception occurred during task execution. To see the full traceback, use -vvv. The error was: If you are using a module and expect the file to exist on the remote, see the remote_src option
failed: [gub37-core] (item=config/dawn.j2) => {
    "ansible_loop_var": "item",
    "changed": false,
    "item": {
        "ctime": 1686774693.5181503,
        "gid": 1000,
        "group": "robertfoss",
        "mode": "0777",
        "mtime": 1686774693.5181503,
        "owner": "robertfoss",
        "path": "config/dawn.j2",
        "root": "/home/robertfoss/projects/freifunk/bbb-configs/roles/cfg_openwrt/templates/corerouter",
        "selevel": "s0",
        "serole": "object_r",
        "setype": "user_home_t",
        "seuser": "unconfined_u",
        "size": 45,
        "src": "roles/cfg_openwrt/templates/ap/config/dawn.j2",
        "state": "link",
        "uid": 1000
    }
}

MSG:

Could not find or access '/home/robertfoss/projects/freifunk/bbb-configs/roles/cfg_openwrt/templates/corerouter/config/dawn.j2' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

PolynomialDivision:
Please try dawn with latest main. There is a very important fix 
https://github.com/openwrt/openwrt/commit/3e1ac00ccbdaa0c396b36429ddbe83d2e3f6276f

PolynomialDivision:
Needs to be backplates to 23.05

pktpls:
Fixed the symlinks and merged as cae0ab090655ae051f42aefec7bed6bf9c880781 & 6dabd1332e8402e20b6cd8b09fbcf9690f642f28 :+1: :cat:

PR #359, opened 2022-10-25
e16: changed frequencies & core router

Noki:


group_vars/model_zyxel_nwa55axe.yml
```diff
@@ -0,0 +1,17 @@
+---
+target: ramips/mt7621
+
+dsa_ports:
+  - lan
```
spolack, line 5: Since this a single port device you dont have to configure the switch. IMO you should better use
```
int_port: lan
```

spolack (dismissed):
Please replace the dsa_ports statement with the int_port. Rest LGTM.

group_vars/model_zyxel_nwa55axe.yml
Noki: will do

group_vars/model_zyxel_nwa55axe.yml
```diff
@@ -0,0 +1,18 @@
+---
+target: ramips/mt7621
+
+compat_version: 1.1
```
spolack, line 4: why do you do this? what does this statement change?

Noki: It's the same fix as here: https://github.com/freifunk-berlin/bbb-configs/commit/0eddf21ab7ed57491d63f38ec6db169f109e6f40

Switching from `dsa_ports` to `int_port` resulted in an upgrade warning.

Noki: for me upgrading without a compat warning only works with compat_version: 1.1 and dsa_ports in combination. I will go with this one therefore.

PR #360, opened 2022-10-25
prometheus: derive our own IP from our config

pktpls:
This has a few advantages:
- we always get the same IP, no matter what the default route is
- we don't need a default route
- we don't require a random internet service to be available


robertfoss:
LGTM

pktpls:
Hm, before merging this and deploying this... this will probably change the `target` for all hosts. I assume we're only using the hostname for graphing anyway?

PR #362, opened 2022-10-25
vm: add vm.sh script for fast rootless x86_64 VMs

pktpls:
Usage: ./vm.sh <location> [<imagedir>]

    location  - Name of a BBB location config file stored at location/<location>.yml.
    imagedir  - Image output directory of BBB-Configs, where *-generic-kernel.img
                and *-generic-ext4-rootfs.img images for <location> can be found.
                Default: /tmp/ansible-openwrt/tmp/imagebuild/<location>-core/bin/targets/x86/64

This script starts a Firecracker micro VM for a given BBB-Configs location's corerouter.
The VM starts very fast, doesn't need root permissions, and gives access to its console.
It exposes SSH/DNS/HTTPS and meshes over a Wireguard tunnel.

This VM is meant for development and debugging, and NOT for permanent deployment.

The VM location's corerouter must have `model: "x86-64"` and it needs registered
IP addresses, just like any regular location. To enable meshing over a tunnel,
you need to configure a network with `role: uplink`, as usual.

The only system requirement is Podman 4.x, since it allows us to operate without root
permissions and without touching the host's network settings. A Podman container with
Alpine Linux is started and its isolated user and network namespaces allow root operations
such as network configuration and VM management.

To create a new VM location, replace "myvm" with your own (see locations/pktpls.yml for a working example).

    vim locations/myvm.yml                          # Create your location config
    ansible-playbook play.yml -l 'myvm-*' -t image  # Build the image
    ./vm.sh myvm                                    # Start the VM

The forwarded ports can be used once the VM has booted ("device eth0 entered promiscuous mode").

    nslookup -port=8053 myvm-core.olsr localhost    # DNS is forwarded
    firefox https://localhost:8443                  # so is HTTPS
    ssh root@localhost -p 8022                      # and SSH

We can even mount the VM's rootfs to work on it directly.

    mkdir vmfs/
    sshfs -v -p 8022 root@localhost:/ vmfs/
    ls -l vmfs/
    umount vmfs/

To shut the VM down, kill it from the outside as Firecracker doesn't support `poweroff`.

    podman kill myvm

Be careful -- all changes are lost when the VM is stopped.

### DIFF
--- a/group_vars/model_x86_64.yml
+++ b/group_vars/model_x86_64.yml
@@ -2,5 +2,8 @@
 
 target: x86/64
 override_target: generic
+image_search_pattern: "*-ext4-combined.img*"
 
 int_port: eth0
+
+wireless_profile: disable
--- a/locations/pktpls.yml
+++ b/locations/pktpls.yml
@@ -1,0 +1,77 @@
+---
+
+location: pktpls
+location_nice: Test von @pktpls
+contact_nickname: Packet Please
+contacts:
+  - pktpls+bbb@systemli.org
+
+hosts:
+
+  - hostname: pktpls-core
+    role: corerouter
+    model: "x86-64"
+
+# feed: "src/gz openwrt_falter file:///home/user/w/ff/falter-packages/out/openwrt-23.05/x86_64/falter"
+# imagebuilder_disable_signature_check: true
+
+location__packages__to_merge:
+  - -luci-mod-falter
+  - -falter-common
+  - -tunnelmanager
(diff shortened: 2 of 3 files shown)


Noki:
The JSCPD linter was disabled with https://github.com/freifunk-berlin/bbb-configs/commit/b6c44839754488cbd7a854ff238bbda53391daf2. Rebase to master should resolve the failed linter checks.

pktpls:
Auto-approved after 389 days

PR #366, opened 2022-11-02
core/uplink: adapt firewall rules to OpenWrt bug fix

PolynomialDivision:
See here:
https://github.com/openwrt/openwrt/commit/ef597b026bb0351ae909ae1fdaed12e76ddd41b7


spolack:
what problem are you trying to solve? where have you seen it?

PR #367, opened 2022-11-02
prometheus: avoid ip --json

pktpls:
Due to space constraints, we don't install `ip-tiny` nor `ip-full` which provide JSON output. Instead we only have Busybox's `ip`, so we have to parse its output on our own.


Noki (approved):
looks good to me. tested on an AP and core-router and outputs the subnet

PR #371, opened 2022-11-04
gruni73: Fix mesh VLANs for uplink antennas

robertfoss:
The mesh vlan entries were mistakenly removed in an earlier commit.

Fixes: 218ffb3 - gruni73: Remove Mikrotik SXTsq 2Ghz APs


pktpls:
:+1:

PR #376, opened 2022-11-10
get rid of compat_version checking

spolack:
this involves hardcoding the resulting compat_version in the image metadata to 1.0 and also hardcoding it within the config. We dont need the compat_version check because we ship the config within the image.

If the device you are flashing is running at a higher compat_version you likely have to use sysupgrade -n in order to fix it again

### DIFF
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -38,5 +38,3 @@ all_sysctl__to_merge:
 
   # especially on low mem devices this is important
   vm.min_free_kbytes: 1024
-
-compat_version: 1.0
--- a/group_vars/model_netgear_wax202.yml
+++ b/group_vars/model_netgear_wax202.yml
@@ -1,8 +1,6 @@
 ---
 target: ramips/mt7621
 
-compat_version: 1.1
-
 dsa_ports:
   - wan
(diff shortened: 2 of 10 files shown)


PolynomialDivision (approved):

Akira25:
@spolack Do you maybe wan't to add the compat-version 42 thing we spoke on in the matrix chat?

spolack:
I just pushed an updated changeset. 

Trying to flash a new bbb-configs image to scharni-ap-keller which has running compat_version: 1.0
```
root@scharni-ap-keller:/tmp# sysupgrade -T scharni-ap-keller.bin 
Fri Nov 11 22:51:00 CET 2022 upgrade: The device is supported, but this image is incompatible for sysupgrade based on the image version (1.0->9.9).
```

After updating compat_version to 9.9 it would update
```
oot@scharni-ap-keller:/tmp# sysupgrade -T scharni-ap-keller.bin 
root@scharni-ap-keller:/tmp# 
```

Akira25:
Looks good to me. I wonder why this wasn't merged before. I remember, that we reached consensus on this. If it was to be other, please feel free to revert then.

PR #377, opened 2022-11-12
Fixup default contact email on community sites

spolack:
Currently it would result in
```
info+zwingli-core@berlin.freifunk.net
```

But we agreed on using `info+$locationname@berlin.freifunk.net`


Akira25:
Looks good to me. I didn't know, that this variable exists.

Akira25 (approved):

PR #382, opened 2022-11-27
kiehlufer: add new ap

freifunkoperator:


group_vars/model_tplink_cpe210_v1.yml
```diff
@@ -11,3 +11,4 @@ wireless_devices:
     htmode_prefix: HT
     path: platform/ahb/18100000.wmac
     ifname_hint: wlan2
+    antenna_gain: 9
```
Noki, line 14: Why do we need this? Driver default works for me just fine:

image

Noki (changes requested):
please remove the antenna gain as stated in my first comment

PR #383, opened 2022-11-29
Dragonkiez more networks

pmelange:
There will be, in the next days, new networks in the dragonerarial.  These commits accomplish this task


pmelange:
I just rebased off of master to remove the new conflict created by pr #384

Akira25 (approved):
Looks good to me. In terms of the adjustment of the ubnt-ac-mesh-model we might consider adding another role for even less flash laster on.

PR #388, opened 2022-12-08
dragonkiez-kiezraum: move dragonkiez-core to dragonkiez-kiezraum

pmelange:
Since there is no specific core router in the dragonkiez, move this setup to dragonkiez-kiezraum


Akira25 (approved):
Looks good to me.

PR #389, opened 2022-12-08
dragonkiez-plangarage: rename dragon-core to dragonkiez-plangarage

pmelange:
Since there are multiple networks in the dragonkietz, it doesn't make sense to have any single one of them to be named core.  Instead rename to it's actual location.


Akira25 (approved):
LGTM.

PR #399, opened 2022-12-22
[DRAFT] Update changes from bbb to wiki.freifunk.net

FFHener:
The goal is to have a tool that updates the wiki since it is often outdatet in comparison to bbb-configs. @Akira25 already wrote a [script](https://github.com/freifunk-berlin/falter-wiki-bot)  to push changes into the wiki. This is still an early draft. It still needs to be discussed which the community if and how an automation coud make sense.

While testing the script we found several locations using a non network adress as a network adress. It still needs to be investigated whether my changes would affect the assigned adresses.

The conecept right now is to extract all relevant data from the yml-files, convert them into the format of the semantic properties from the wiki and than have in every wikipage that should be updatet a section with all the values hidden. We can than point in the article to these values and just update the values without touching the articles. This enables a big flexibility. It would also make it possible to extract these data from the wiki. [Here](https://wiki.freifunk.net/Spezial:Durchsuchen?title=Spezial%3ADurchsuchen&article=Berlin%3AStandorte%3ASamariterkirche) is an example from Sama for the semantic properties.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 */__pycache__/
 .vscode/
 venv/
+.env
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@
 
 BBB-Configs manages and provisions OpenWrt mesh nodes in the city-wide backbone of Freifunk Berlin via ansible. It abstracts the typical OpenWrt mesh config to generic templates. Those templates provision all locations more or less the same helping the maintainers to orientate on all sites.
 
-The provision of a router works by generating the necessary OpenWrt configs and feeding the OpenWrt-Imagebuilder with it. In the end, ansible generates a self-contained binary image ready to be flashed on a router. Due to the self-contained property bbb-configs can ensure to function properly on a router in a specific location. Maintainers can remotely upgrade sites without having to worry about wrong configurations.
+The provision of a router works by generating the necessary OpenWrt configs and feeding the OpenWrt-Imagebuilder with it. In the end, ansible generates a self-contained binary image ready to be flashed on a router. Due to the self-contained property bbb-configs can ensure to function properly on a router in a specific location.
+Maintainers can remotely upgrade sites without having to worry about wrong configurations.
 
 ## Getting Started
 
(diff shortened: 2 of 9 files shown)


Akira25:
Thank you for you draft on implementing this. Really appreciated. 

I especially like the idea of using sematic Wiki, which is a kind of database in a mediawiki-installation. (further readings [here](https://www.semantic-mediawiki.org/wiki/Semantic_MediaWiki) and [here](https://www.semantic-mediawiki.org/wiki/Demo:Examples/Queries)). This enables us with a whole lot of opportunities in using this data.

Though I wonder a bit why you chose the way of parsing the yaml-files by yourself. Ansible already does this for us. You then can use the parsed variables directly in so called templates to render them into correctly formatted text. This might be a more straight forward approach, which can save a lot lines of code.

For a simple example of a template you might have a look at the roles directory. The template for the uci system configuration might be a good starting point, as it is not too complex and shows a lot of possible structures:
https://github.com/freifunk-berlin/bbb-configs/blob/master/roles/cfg_openwrt/templates/common/config/system.j2 

The templating language is called Jinja2.

One notice on the `__pycache__`-dir. It would be cool to not have it in this PR, but to keep it as your local file. :)

For the irregularities you've found in the data, we should enforce a standard. If some data is missing, we shouldn't try to add it magically, but kindly ask the respective maintainer to provide it for us. That will keep things more simple and additionally help keeping things maintainable. For example we maybe should consider failing hard in yml2wiki.py:217-222 instead of just adding information, that isn't given.

Really cool that you've seen the issues with the prefixes. Indeed it should be the networks address itself and not the one of  the first device in the network. As I expect ansible to handle this intelligently, it shouldn't do any harm though. Maybe we can ask @spolack for his assessment on this.

FFHener:
> Really cool that you've seen the issues with the prefixes. Indeed it should be the networks address itself and not the one of the first device in the network. As I expect ansible to handle this intelligently, it shouldn't do any harm though. Maybe we can ask @spolack for his assessment on this.

Done some digging in my own documentation. In october i made 2 config runs with the original adresses and the new adresses. Both times the values in etc/dhcp where the same, so i think it should still be the same.

> Though I wonder a bit why you chose the way of parsing the yaml-files by yourself. Ansible already does this for us. You then can use the parsed variables directly in so called templates to render them into correctly formatted text. This might be a more straight forward approach, which can save a lot lines of code.

Yes. You are right, thank you for pointing out (again) to me. Ansible was a big black box for me so I just started to build on top of ideesnoires work. I now took the time, it will be way cleaner.

FFHener:
@Akira25 Would you be up for checking out the current state and give your feedback? It is mostly in a state where i would say it could be merged as v1, just some linterstuff and one error message needs improvement in my opinion.

Akira25 (approved):
Neat. I like it.

robertfoss:
Looks good to me

FFHener:
I found 2 minor things while doing more testing with the tool. We need the link between Wikiarticle and Node, so for a newly created article one has to add the wikiupdater-output manually once or add the coordinates by hand. In Addition Mediawiki interprets a ``#bar:matrix.org`` as a [URI fragment](https://en.wikipedia.org/wiki/URI_fragment) resulting in ``Berlin:Standorte:Foo#bar:matrix.org``

noxilixon:
The first problem should be solved with a clear documentation. The second one could be solved by using a `https://matrix.to/` link, which might be handy anyway. Best way would be to detect this with regex (something like `#.*\:.*\..*`) and then add the link automatically.

FFHener:
I did again a configrun with all locations that completed without any errors (leaving huette15-core aside because it's using a non-standard wireguard config). Since all linting errors are fixed I'm happy for more reviews or going into "production" with it. The list of articles already converted to this as an example are:
* fesev
* magda
* l5

The [LOS-Visualizer](https://wiki.freifunk.net/Berlin:Line-of-Sight_visualiser) got a [rewrite](https://github.com/freifunk-berlin/util.berlin.freifunk.net/pull/16) and will continue to work with the converted articles.

robertfoss:
> I tried to reproduce your error, but I couldn't. I used the latest wiki updater version with the latest gub37 version. My ansible version is `ansible-playbook [core 2.14.5]`.

```
$ ansible --version
ansible [core 2.14.3]
```

FFHener:
@robertfoss and I couldn't reproduce the error in pm, so he gave his approval to merge this PR.

Akira25:
Glad this is in master now. Thank you very much for this contribution! :)

PR #405, opened 2023-01-19
ak36-gw: add new radio to rhnk

pmelange:
From as36s-hyp03 to the rooftop switch, mesh is on VLAN 14

mgmt ip 10.31.130.136
mesh ip 10.31.130.165 2001:bf7:750:4001::6/128


Noki (approved):
looks good to me

PR #406, opened 2023-02-07
Two *.olsr domain name fixes

pktpls:
dnsmasq: define .olsr domains independently of OLSRd

> Previously, dnsmasq would only pick up our own .olsr domain names
> once OLSR had first flushed its nameservice hosts file, and once
> dnsmasq actually reloaded its config. This could take up to 5 minutes.
> Until then, our own .olsr domains were not resolvable.
> 
> We now explicitly tell dnsmasq about our own .olsr domain names.
> As a result, they can be resolved instantly once dnsmasq is running.
> 
> Neat side effect: proper .olsr hostnames in Reverse DNS and traceroute.
> Before this patch, the first hop would only report as 'frei.funk'.

dnsmasq: never forward .olsr queries to public resolvers

> Queries for *.olsr should never be forwarded to public resolvers.
> Some resolvers drop queries for unknown TLDs, without a response.
> 
> As a result, our query waits for a long time and eventually times out.
> This affects A queries for unknown hosts, and also all AAAA queries.
> 
> Happy Eyeballs means an AAAA query is sent almost always,
> even though *.olsr is really only used for our IPv4 networking,
> and so we get lots of very slow DNS queries...
> 
> This can be avoided very simply by not forwarding .olsr queries.

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
+++ b/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
@@ -18,6 +18,15 @@ config dnsmasq
         list server '{{ _dns_server }}'
 {% endfor %}
 
+{% for network in networks | selectattr('role', 'equalto', 'mgmt') %}
+  {% for host, ip_num in network['assignments'].items() %}
+config domain '{{ host | replace('-', '_') }}_olsr'
+	option name '{{ host }}.olsr'
+	option ip '{{ network['prefix'] | ansible.utils.ipaddr(ip_num) | ansible.utils.ipaddr('address') }}'
+
+  {% endfor %}
+{% endfor %}
+
 {% for network in networks | rejectattr('role', 'in', ['ext', 'mesh']) %}
     {% set name = network['name'] if 'name' in network else network['role'] %}
 


roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
```diff
@@ -13,10 +13,20 @@ config dnsmasq
...
 {% endfor %}
 
+{% for network in networks | selectattr('role', 'equalto', 'mgmt') %}
+  {% for host, ip_num in network['assignments'].items() %}
+config domain '{{ host | replace('-', '_') }}_olsr'
+	option name '{{ host }}.olsr'
+	option ip '{{ network['prefix'] | ansible.utils.ipaddr(ip_num) | ansible.utils.ipaddr('address') }}'
+
+  {% endfor %}
+{% endfor %}
```
Akira25, line 28: Do I understand this right, that this will include a list of all bbb-configs-known nodes to the dnsmasq config?

roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
```diff
@@ -13,6 +13,7 @@ config dnsmasq
 	option noping '{{ dhcp_no_ping|int if dhcp_no_ping is defined else 1 }}'
 	option quietdhcp '1'
 	option cachesize '{{ dns_cache_size | default(10000) }}'
+ 	list server '/olsr/'
```
PolynomialDivision, line 16: Can you explain the config syntax? I don't understand why this config option should exclude olsr domains from a lookup.

PolynomialDivision: I found it: https://openwrt.org/docs/guide-user/base-system/dhcp_configuration#selective_dns_forwarding

Probably you could also do `list server '/olsr/127.0.0.1'`

pktpls: > Probably you could also do list server '/olsr/127.0.0.1'

That *might* do a loop, `/olsr/` simply means that *.olsr lookups are forwarded to a "null server" if we don't already know the answer, i.e. they're not forwarded at all.

roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
pktpls: Only all mgmt assignments of the current location, as in `networks` in networks.yml. It's basically the same listing that we do in olsrd.conf to configure the nameservice plugin.

Making this explicit to dnsmasq means we don't need to wait precious minutes for the domain names to become available.

roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
pktpls: If OLSRd served a location DNS server on e.g. port 10053 instead of writing files, we could set it to `/olsr/127.0.0.1#10051` instead of refreshing files.

Noki: > If OLSRd served a local DNS server on e.g. port 10053 instead of writing files, we could set it to `/olsr/127.0.0.1#10051` instead of refreshing files.

That would probably solve https://github.com/OLSR/olsrd/issues/116 along the way.

PR #410, opened 2023-02-12
corerouter: instead of domains use hostrecords

PolynomialDivision:
We actually talk about hosts and not domains. So lets switch to hostrecords.

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
+++ b/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
@@ -20,9 +20,9 @@ config dnsmasq
 
 {% for network in networks | selectattr('role', 'equalto', 'mgmt') %}
   {% for host, ip_num in network['assignments'].items() %}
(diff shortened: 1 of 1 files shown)


pktpls:
What is the difference?

Noki:
> What is the difference?

Between a domain and a host? The domain includes the domain and all subdomains within, the host is only a single host. So a domain example.com usually includes "example.com", and "*.example.com", but the host is only that specific host.

Noki:
Maybe it would also be a good idea to use subdomains for the different devices in a core-router setup like:

example-core.olsr - corerouter
ap1.example-core.olsr - accesspoint

When registering IPs the name is considered uniqu, so example-core will be kind of protected, but all the other hosts aren't registered and could therefore exist twice in the network.

pktpls:
I'm open for anything but I still don't understand the change in this PR :-)

Is it that we currently define entire zones in dnsmasq, while we actually only need individual host entries similar to what's in /etc/hosts?

PR #412, opened 2023-02-14
all: fix package conflicts on snapshot

Noki:
This fixes conflicts of package dependencies which broke snapshot builds.


PolynomialDivision:
Fine for me, but maybe we should change it in the future to mbedtls. It's now OpenWrt default and probably it is better to stay with upstream solutions.

PR #415, opened 2023-02-20
MT7915E: switch to snapshot

Noki:
Snapshot uses the newer 5.15 Linux kernel. By switching the devices to snapshot the issues with unstable mesh connections are now gone.

Note: I did not include the ZyXEL NWA50AX into this commit, as it is seems to be used only as AP, however it will need to be switched to snapshot if someone wants to use it for meshing.


PolynomialDivision:
can you rebase on latest master?

PR #422, opened 2023-03-06
fritzbox7530: add dsa variant for snapshot

pktpls:
DSA conversion is neccessary to run OpenWrt snapshot so that we can test the VDSL modem.

Switch your host's `model` to `avm_fritzbox-7530_dsa` to try it out.

### DIFF
--- a/group_vars/model_avm_fritzbox_7530_dsa.yml
+++ b/group_vars/model_avm_fritzbox_7530_dsa.yml
@@ -1,0 +1,26 @@
+---
+openwrt_version: snapshot
+target: ipq40xx/generic
+override_target: "avm_fritzbox-7530"
+
+model__packages__to_merge:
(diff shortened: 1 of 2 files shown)


group_vars/model_avm_fritzbox_7530_dsa.yml
```diff
@@ -0,0 +1,27 @@
...
+model__packages__to_merge:
+  - "-kmod-ath10k-ct -ath10k-firmware-qca4019-ct"
+  - "kmod-ath10k ath10k-firmware-qca4019"
+
+dsa_ports:
+  - lan1
+  - lan2
+  - lan3
+  - lan4
+  - wan
```
PolynomialDivision, line 15: the 7530 does not have a wan port? or are you talking about the dsl line?

pktpls: Oops, you're right, I also removed the wan port from the old model file in a second commit.

PR #425, opened 2023-03-07
wireless: remove owe transition

freifunkoperator:
OWE is buggy. Remove it until we were able to fix everything.

### DIFF
--- a/group_vars/all/wireless_profiles.yml
+++ b/group_vars/all/wireless_profiles.yml
@@ -49,44 +49,6 @@ wireless_profiles:
         network: dhcp
         radio: [11a_standard, 11g_standard]
         ifname_hint: ff
-        owe_transition_ifname_hint: ffowe
-
-      - mode: ap
-        ssid: berlin.freifunk.net OWE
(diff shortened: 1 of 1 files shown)


pktpls (approved):
Thanks!

pktpls:
For reference, there's a bug in wpa_supplicant or hostapd or OpenWrt's hostapd patches, which breaks the 4th step of the 4-way-handshake: `received EAPOL-Key msg 4/4 in invalid state (7) - dropped`. Attempts of connecting to an SSID pair with OWE Transition Mode will fail in most cases (though not *all* cases).

What we guessed earlier, is that interop between old Open-only SSIDs at some locations and new Open+OWE SSIDs at other locations was the issue, but that is *not* the case. Interop is fine and the handling of both types together in client implementations is also mostly fine.

We probably are dealing purely with an implementation bug of the WPA3 handshake in a Transition Mode environment. wpa_supplicant's latest main branch doesn't seem to have the bug, so there are some cues that'll help with finding and fixing it.

PR #430, opened 2023-03-14
Added defaults for switch_ignore_ports

FFHener:
Empty switch_ignore_ports-lists are no longer required in the model files. Removed it from the existing model-files and updated docs.


Noki (approved):
looks good to me

Issue #416, opened 2023-02-21
Lint errors after ansible update


noxilixon:
When I reinstalled the requirements I got `ansible-lint 6.13.1 using ansible 2.14.2` as ansible version. This matches the version defined in the requirements.txt:
```
...
ansible >= 2.13
ansible_merge_vars >= 5.0.0
...
```
But this version seems to use new linting rules. When executing the lint with this version I get the following errors:

```
> make lint
pip3 install -q -r requirements.txt
yamllint -d .config/yaml-lint.yml .
ansible-lint -c .config/ansible-lint.yml
WARNING  Listing 22 violation(s) that are fatal

name[play]: All plays should be named.
play.yml:2

risky-file-permissions: File permissions unset or incorrect.
roles/cfg_openwrt/tasks/add_symlinks.yml:2 Task/Handler: Create rc.d directory

ignore-errors: Use failed_when and specify error conditions instead of using ignore_errors.
roles/cfg_openwrt/tasks/fix_permissions.yml:12 Task/Handler: Set permission for prom-catalog-register script

ignore-errors: Use failed_when and specify error conditions instead of using ignore_errors.
roles/cfg_openwrt/tasks/fix_permissions.yml:18 Task/Handler: Set permission for replace_secrets script

risky-file-permissions: File permissions unset or incorrect.
roles/cfg_openwrt/tasks/imagebuilder.yml:2 Task/Handler: Create images dir

risky-file-permissions: File permissions unset or incorrect.
roles/cfg_openwrt/tasks/imagebuilder.yml:16 Task/Handler: Create Download directory

risky-file-permissions: File permissions unset or incorrect.
roles/cfg_openwrt/tasks/imagebuilder.yml:26 Task/Handler: Download Imagebuilder if upstream is newer

no-changed-when: Commands should not change things if nothing needs doing. (warning)
roles/cfg_openwrt/tasks/imagebuilder.yml:32 Task/Handler: Copy Local Imagebuilder

risky-file-permissions: File permissions unset or incorrect.
roles/cfg_openwrt/tasks/imagebuilder.yml:45 Task/Handler: (Re)-Create workdir

no-changed-when: Commands should not change things if nothing needs doing. (warning)
roles/cfg_openwrt/tasks/imagebuilder.yml:56 Task/Handler: Read instruction set

risky-file-permissions: File permissions unset or incorrect.
[...]

PR #431, opened 2023-03-14
Fixed liniting errors for latest ansible-lint

noxilixon:
Fixed the errors and warnings described in #416.

- I have defined the permissions to the files, according to the permissions in the configs created with bbb-configs on my machine.
- Commented out the "ignore_errors" in two tasks, because I wanted to fix [the warning](https://ansible-lint.readthedocs.io/rules/ignore-errors/) but didn't know what errors could rise and got none when creating all configs by myself.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ ansible_merge_vars >= 5.0.0
 jmespath >= 1.0.1
 netaddr >= 0.8.0
 ansible-lint >= 6.5.2
-yamllint >= 1.27.1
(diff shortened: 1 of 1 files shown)


Noki (changes requested):
Please remove the comment lines completely. Please also update the requirements.txt to the latest version of ansible-lint (`ansible-lint >= 6.14.3`) and change the commit message to start with "ansible-lint:" e.g. "ansible-lint: update ansible-lint and fix corresponding liniting errors"

Issue #120, opened 2021-12-20
Implement Switchport Profiles


spolack:
**RFC**

A switchport profile should be a hash of port config, which can be assigned to all or individual ports per device.
```
port_profile_name:
  networks: [network_name, ..]
  native_network: network_name
  poe_mode: type poe_mode

type poe_mode: [24_passive, 48_passive, af]
```

Following Profiles should be implicitly created. profile all is applied by default to every port

```
all:
  networks: [all networks]
  native_network: None
  poe_mode: af

all_24_passive:
  networks: [all networks]
  native_network: None
  poe_mode: 24_passive

{per network with vid}
$network_name:
  networks: [all networks]
  native_network: $network_name
  poe_mode: af
{end}

{per network with vid}
$network_name + "_24_passive":
  networks: [all networks]
  native_network: $network_name
  poe_mode: 24_passive
{end}

{per network with vid}
$network_name + "_mgmt":
  networks: [$network_name, mgmt]
  native_network: $network_name
  poe_mode: af
{end}

{per network with vid}
$network_name + "_mgmt_24_passive":
  networks: [$network_name, mgmt]
  native_network: $network_name
  poe_mode: 24_passive
{end}
```

PolynomialDivision:
Can you give a specif example, how a config would look like, for a site? I do not understand exactly each line of the template code you wrote.

Noki:
Just wanted to point out two usecases:

1. In a setup with a core router and multiple accesspoints that all have multiple LAN ports and a WAN port the WAN port could be used to connect the AP to the core router. All other LAN ports could be configured to use VLAN 40 so users could directly connect to the APs via cable and without having to deal with a VLAN configuration. This is a well known setup to most users. This could also be setup as private VLAN so all devices connected via LAN to an AP (pc, printer, nas, tv, ...) could communicate with each other but are isolated from the rest of the setup.
2. Devices like a webcam or server that might not have or support VLAN tagging could be connected to a seperate LAN port that is assigned to a VLAN without filtering so it could be made accessible via Freifunk and the internet.

spolack:
I just gave some different techniques to prepopulate the profiles a try and found a way to use loops in templates which doesnt feel too hackish. Not sure if they are better ways though.

Commit:
https://github.com/freifunk-berlin/bbb-configs/commit/d2de94a4375b3e90aa1798456330912b24ce71fc


Resulting port_profiles variable:

```
all:
  native_vlan: false
[...]

PR #451, opened 2023-04-12
[WIP, will not be merged] roles: add possibility to untag ports for certain hosts

PolynomialDivision:
You can use this by adding to the host file:

	port_untag:
	  40: [2,3,4]

### DIFF
--- a/roles/cfg_openwrt/templates/common/config/dsa.network.inc
+++ b/roles/cfg_openwrt/templates/common/config/dsa.network.inc
@@ -6,6 +6,9 @@ config device
   {% set portmapping = [] %}
   {% for port in dsa_ports %}
     {% set tagged = not network.get('untagged') %}
+    {% if port_untag is defined and network['vid'] in port_untag and port in port_untag[network['vid']] %}
+    {% set tagged = false %}
+    {% endif %}
     {{ portmapping.append(port|string + (":t" if tagged else "")) }}
   {%- endfor %}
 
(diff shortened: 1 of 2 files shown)


Noki:
This is great! Does the / can we make the following syntax work?

```
port_untag:
  13: [1]           # untag the wireguard uplink vlan - corerouter is plugged into a normal home router
  40: [2,3,4]     # untag the default dhcp network - clients can get a dhcp address
  41: [5]          # untag a second dhcp network - network used for hosting services without client isolation / inbound filtering
```

I think the untag option on the networks would then have to go.

Noki:
Maybe it would be even better to pull all the port related stuff into a ports section as another idea on how to structure switchport profiles (https://github.com/freifunk-berlin/bbb-configs/issues/120).

```
ports:
  vlans:
    untagged:
      13: [1]             # untag the wireguard uplink vlan - corerouter is plugged into a normal home router
      40: [2,3,4]         # untag the default dhcp network - clients can get a dhcp address
      41: [5]             # untag a second dhcp network - network used for hosting services without client isolation / inbound filtering
    tagged:
      21: [6]             # vlan for antenna
      22: [7]             # vlan for antenna
      42: [1,2,3,4,5,6]   # management vlan
  poe: 
    on: [1,2,3,4]         # enable poe on ports 1-4
    pow_mode:
      24_passive: [1,2]   # set poe_mode to 24_passive on ports 1-2
      af: [3, 4]          # set poe_mode toaf on ports 3-4
  ```

**Comment by pktpls** [comment]
I have a feeling the simplest way is to stay close to OpenWrt's config:
```yml
    ports:
      mesh_core: lan1
      dhcp: lan2
      privdhcp: lan3
      mgmt: 'lan1:t lan2:t lan3:t lan4 wan:t' # untagged only on lan4
      vdsl: wan
```

PR #455, opened 2023-04-23
templates: naywatch: don't save logs

PolynomialDivision:


pktpls:
Yes, nice :+1: 

For the record: some setups trigger naywatch very often and accumulate a ton of (useless) logs, and the logs are even carried over by sysupgrade, so they eventually run out of overlay diskspace.

PR #463, opened 2023-05-18
DEVELOPER.md: Updatet VLAN convention in regards to wireguardtunnel

FFHener:


FFHener:
I'm not to shure why Ansbible Test build even got triggered to build three locations, but since the error-message (see below) has nothing to do with my changes I think this PR could be merged. This maybe means more work has to go into #485. In addition I fixed some yml-linting errors that occured in ``developer.md``.

```
Configuring collectd-mCollected errors:
 * pkg_hash_check_unresolved: cannot find dependency libubox20230523 for mtd
 * pkg_hash_fetch_best_installation_candidate: Packages for mtd found, but incompatible with the architectures configured
 * opkg_install_cmd: Cannot install package mtd.
 ```

Noki:
We always build 3 locations, even if no location was changed.

Issue #426, opened 2023-03-07
OWE Transition Mode


pktpls:
In the past couple of months we tried out OWE Transition Mode on the "berlin.freifunk.net" SSID. This has been rolled back because of a bug:

pktpls:
In the meantime @PolynomialDivision has updated OpenWrt hostapd (openwrt/openwrt#12178), but it looks like the EAPOL-key bug persists.

Which makes me think that one of Openwrt's patches to hostapd is responsible. Primary candidates would be the cryptography-related and wifi-related patches, I guess: https://github.com/openwrt/openwrt/tree/main/package/network/services/hostapd/patches

pktpls:
It works with OpenSSL :+1: 

With WolfSSL it's broken :-1: some connection attempts with the aforementioned `EAPOL-Key` error from hostapd.

With Mbed TLS it's broken too :-1: all connection attempts fail, but without any error message from hostapd.

PolynomialDivision:
I think in comparison openssl is taking way more flash? I think that's why we currently stay with wolfssl or mbedtls?

pktpls:
Not reproducible with WolfSSL on current snapshot mt76 / zyxel_nwa50ax, I'll also have to try that Fritzbox again

pktpls:
Bug still triggers with WolfSSL and current OpenWrt snapshot on ipq40xx / avm_fritzbox-4040

pktpls:
Update:

The issues with WolfSSL and MbedTLS are specific to ipq40xx and have been reproduced on avm_fritzbox-4040 and mikrotik_wap-ac.

We can easily work around it by installing OpenSSL in our ipq40xx target file.

pktpls:
Another fixable OWE issue:

The Broadcom `bcmwl` driver on Linux can't deal with OpenWrt's default `wpa_key_mgmt` settings: https://bugs.launchpad.net/ubuntu/+source/bcmwl/+bug/1956162 - I reproduced this with my friend's old Fedora Macbook Air.

We should test setting `wpa_key_mgmt=WPA-PSK SAE`.

pktpls:
iOS now does OWE but only with devices 2021++ https://support.apple.com/en-gb/guide/deployment/dep3b0448c58/web

> iOS 16, iPadOS 16.1 and macOS 13, or later, support a way to secure open Wi-Fi networks, using the standard Opportunistic Wireless Encryption. This standard is supported on the following devices:

PR #466, opened 2023-05-21
wireless_profiles: OWE requires mandatory managment frame protection

PolynomialDivision:


Noki:
@PolynomialDivision https://github.com/freifunk-berlin/bbb-configs/pull/467 this disables the JSCPD linter, because we have duplication by design.

pktpls:
I don't understand, I thought Transition Mode requires mandatory 80211w. OWE itself has been working fine from what I can tell

Noki (approved):
> I don't understand, I thought Transition Mode requires mandatory 80211w. OWE itself has been working fine from what I can tell

As far as I understand it, setting it to be optional `1` is/was the problem. Setting it to `2` is correct as it is mandatory. So this looks good to me.

pktpls:
This doesn't fix Transition Mode (#426), I just tried 23.05-SNAPSHOT in Fritzbox 4040 and hit the same handshake bug: `received EAPOL-Key msg 4/4 in invalid state`.

The reason we switched 80211w from 2 to 1 in the past is that it can cause problems with 80211r roaming: https://github.com/openwrt/openwrt/issues/9181

Does this PR fix OWE for iOS? What does it fix? :-)

pktpls:
Works fine without management frame protection :+1: see #426 our issue was some ipq40xx-specific thing which we can work around by using OpenSSL

PR #485, opened 2023-06-09
workflow: make the detection of changed hosts and locations more robust

Noki:
Before this a change of e16 caused a build of e16outdoor as well because a wildcard was used for the location. Now everything is more precise and duplicates also get removed.


Akira25 (approved):
LGTM.

PR #486, opened 2023-06-10
rhxb: add dragon

pmelange:
There are 6 clients which currently use connections which are poor quality.  This adds a new dish which will serve the dragonerareal

### DIFF
--- a/group_vars/location_rhxb/networks.yml
+++ b/group_vars/location_rhxb/networks.yml
@@ -64,35 +64,41 @@ networks:
     ipv6_subprefix: -7
     ptp: true
 
+  - vid: 17
+    role: mesh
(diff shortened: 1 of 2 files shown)


group_vars/location_rhxb/networks.yml
```diff
@@ -64,35 +64,41 @@ networks:
     ipv6_subprefix: -7
     ptp: true
 
+  - vid: 17
+    role: mesh
+    name: mesh_dragon
+    prefix: 10.31.126.135/32
```
Noki, line 70: This IP is already used for mesh_no_5ghz

pmelange: Thanks.  Fixed

PR #488, opened 2023-06-14
dragonkiez-adlerhalle: fix snmp setup

pmelange:
small but import change for the statistics

### DIFF
--- a/group_vars/location_dragonkiez_adlerhalle/snmp.yml
+++ b/group_vars/location_dragonkiez_adlerhalle/snmp.yml
@@ -1,6 +1,6 @@
 ---
 
 snmp_devices:
-  - hostname: dragonkiez-dorfplatz-rhxb
+  - hostname: dragonkiez-adlerhalle-rhxb
     address: 10.31.34.46
     snmp_profile: airos_8


Noki:
lgtm

pmelange:
@Noki why did you force push the branch?  The change made by commit d86cf90dfbaedb3414ca3ce6205ace94974c9a97 does not conflict with the changes made on branch dragon-adlerhalle-snmp and so it is unclear why the force push was done.

pktpls:
Build failure is unrelated to this PR (go merge) and seems to be caused by squashfs image size exceeding partition size.

https://openwrt.org/toh/ubiquiti/unifiac

```
...
7428906 bytes (7.4 MB, 7.1 MiB) copied, 0.0122586 s, 606 MB/s
padding image to 00950000
bash: line 1: 8: Bad file descriptor
...
```

Size 0x950000 is 9764864 bytes, while the partition is only 7929856 bytes (size 0x790000). The rootfs contents need to be reduced until squashfs chooses a padding step that results in less that 7929856 bytes image size.

Noki:
@pmelange I rebased, because I thought this would fix the test build error. Merged now...

PR #496, opened 2023-07-04
Rename uplink_gateway => gateway, ext => uplink

pktpls:
Little renaming just for simplicity. This makes it clearer that `gateway` host role and `uplink` network role are separate things, which will help us communicate in the future. (Also `ext` was a pretty meaningless name).

The inventory thing is a little startup performance optimization, `ansible-inventory --list` is roughly 10 % faster.


robertfoss:
LGTM

PR #497, opened 2023-07-05
inventory: support single-file locations

pktpls:
This adds support for locations defined in a single file, in addition to existing locations defined in Ansible's default host_vars/group_vars structure.

```yml
# Example location file in locations/pktpls.yml
---
location: pktpls
var: value-from-location
hosts:
  - hostname: pktpls-core
    role: corerouter
    var: value-from-host
```
To construct the host vars, the `hosts` array is removed from the location object. Then the host object with a matching hostname is merged into the location object. The result is:
```yml
---
hostname: pktpls-core
role: corerouter
var: value-from-host
location: pktpls
```

I wasn't expecting the implementation to be so simple, or to work in parallel with the host_vars/group_vars way - but here we are :)


Noki:
I like that this eliminates host_vars but what about custom wireless profiles? Can we add those as well?

pktpls:
> I like that this eliminates host_vars but what about custom wireless profiles? Can we add those as well?

Next step :-) SNMP as well

pktpls:
Now also supports wireless_profiles right in the location file, by using merge_vars.

robertfoss:
I like this

PR #501, opened 2023-07-10
all: simplify how SNMP config is rendered

pktpls:
There's no need to generate the <Data> blocks, it's always the same for every host. It also adds a lot of noise to each host's inventory data.

Keep all this static data right in template files, which is much simpler.


spolack:
This breaks with the (not hardly enforced) paradigm to seperate code/templates from the actual configuration. What is the problem with "noise" by adding common values to the inventory?

pktpls:
What exactly is config data can be debatable, there's plenty of hardcoded values in config templates which is good because they never change and are the same for all hosts. The nature of data depends on perspective and I feel like from a location's perspective it doesn't matter what the OIDs and types of all metrics are.

I often look at Ansible's inventory data for debugging and always notice that ~50% of the stuff in there is noise (counted by lines), with SNMP accounting for ~15% and wireless_profiles ~35%. For many hosts, that data isn't even used, and for rest, only a small fraction is used.

I can't tell if smaller inventories also have a perf benefit, if they do it's negligible.

Another nice alternative would be to define SNMP targets as regular hosts, which would pave the way for integrating their configs into bbb-configs in the future.

I'm just looking for easy wins to make bbb-configs' data model have a clean surface. This particular bit is not terribly important for me, since the SNMP profile data is never mentioned in host/location configs. wireless_profiles seems much more impactful :)

FFHener:
> Another nice alternative would be to define SNMP targets as regular hosts, which would pave the way for integrating their configs into bbb-configs in the future.

Personally I do like this approach more as it opens more options up in the future. It would help as well with the perspective of fully documenting Backbone Infrastructure here in git.

PR #504, opened 2023-07-12
files: guard against new falter-common uci-defaults scripts

pktpls:
Now that most uci-defaults script have moved into the falter-common package, bbb-configs needs to be able to disable all of them.

Also see https://github.com/freifunk-berlin/falter-packages/pull/395


robertfoss:
Thanks for looking into this, it's been pretty darn annoying.

Works: https://10.31.157.1/cgi-bin/luci/admin/status/olsr/neighbors
Broken still: https://10.31.157.1/cgi-bin/luci/admin

pktpls:
This is not supposed to fix all the issues on master. It only prevents our config files from being overwritten :)

robertfoss:
LGTM

PR #505, opened 2023-07-13
network: randomize MAC address of mesh interfaces

robertfoss:
When radios that bridge all VLANs are used in conjunction with unmanaged switches, a situation can arise where the switch gets confused about which port specific MAC address is available on since all of the core router mesh interfaces use the same MAC address.

For example:
The issue happens for me because i'm using a unmanged dumb switch to supply my antennas.

This dumb switch is really bridging all vlans on all ports. Not nice, but that means that mesh Traffic from my wilgu10-sama antenna is bridged to the AF60-LR towards sama, where it gets dropped, because that VLAN is not configured on that port

on linux you have by default the same mac address on every VLAN interface

When following packet is sent it will break it every time Sama-core VLAN 10 ->
sama-switch ->
sama-sued-5ghz (untag vlan10) ->
wilgu10-sama (tag vlan 12) ->
dumb switch (floods broadcast on all ports) ->
wilgu10-sama-60ghz

The wilgu10-sama-60ghz seems to learn the mac address of sama-core router on the local ethernet interface, leading to packet loss. When the next packet from sama is received via 60G its resolved until next packet via 5Ghz/Ethernet is received

### DIFF
--- a/roles/cfg_openwrt/templates/common/config/network.j2
+++ b/roles/cfg_openwrt/templates/common/config/network.j2
@@ -33,9 +33,9 @@ config interface '{{ name }}'
     {% if port_needed %}
 	option device '{{ ('br-' + name) if bridge_needed else port }}'
     {% endif %}
+	option macaddr '22:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}'
     {% if network.get('enforce_client_isolation') and role == 'corerouter' and
           not bridge_needed %}
-	option macaddr '02:00:00:00:00:01'
     {% endif %}
     {% if 'assignments' in network and inventory_hostname in network['assignments'] %}
 	option proto 'static'


Noki:
I think https://docs.ansible.com/ansible/latest/collections/community/general/random_mac_filter.html is what should be used here. It comes with a string prefix and a seed to get repeatable results.

robertfoss:
@Noki I saw that collection, but I couldn't find a way to permanently enable it in our bbb-configs repo. Instead it seems like it needs to be installed externally to the repo, which complicates the build procedure.

Maybe there is a way to host the community collection inside of this git repo though. I just don't know how.

robertfoss:
If the choice is between having a hacky-less-than-ideal one liner, vs introducing and external dependency, I would choose the hacky one liner.

That being said, maybe there is a better way to add this support.

pktpls:
The approach looks fine to me, if you wanna make it nicer then add variable declarations. Otherweise LGTM :+1: 

I feel like we should avoid galaxy collections whenever possible - I really don't understand why they added an additional package management tool :(

roles/cfg_openwrt/templates/common/config/network.j2
```diff
@@ -33,9 +33,9 @@ config interface '{{ name }}'
     {% if port_needed %}
 	option device '{{ ('br-' + name) if bridge_needed else port }}'
     {% endif %}
+	option macaddr '22:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}:{{ lookup('password', '/dev/null chars="digits" length=2') | lower }}'
     {% if network.get('enforce_client_isolation') and role == 'corerouter' and
           not bridge_needed %}
-	option macaddr '02:00:00:00:00:01'
```
PolynomialDivision, line 38: Isn't that breaking client isolation?

robertfoss: Removing this line wasn't intentional. However...

This line also sets `option macaddr` , which conflicts with the randomization. 

Two things:
 1) I don't understand how setting a macaddr enforces client isolation
 2) This set below snippet evaluates to true
 ```
   {% set name = network['name'] if 'name' in network else network['role'] %}
   {% set bridge_needed = name in wifi_networks or network.get('mesh_ap') == inventory_hostname or (role == 'corerouter' and 'tunnel_wan_ip' in network) %}
   {% if network.get('enforce_client_isolation') and role == 'corerouter' and  not bridge_needed %}
  option macaddr '02:00:00:00:00:01'
   {% endif %}
```

1) I'm sure there's an answer to. Either way we can maintain this logic.

2) Is kind of convoluted logic. Relying on having macaddr randomization, which may or may not be enabled due to very convoluted logic is sure to create bugs for us down the line.

Because of 2) maybe we should drop this idea and force radios that bridge all VLANs to live behind a managed switch (as they probably should be doing anyway).

robertfoss:
Closing this, don't think there's any interest,

PR #507, opened 2023-07-13
Bht airos dfs reset

pmelange:
Add airos_dfs_reset and update mesh metrics

This image is currently running on the node.


Noki:
Looks good to me. However no need but also no harm to set mesh_metric to 512, which is the default anyway.

pmelange:
Oh, I wanted a "worse" metric for this link.  I guess 1024.  I will create a new PR later to do that.

PR #516, opened 2023-07-24
bht: modify the dfs timing and metric at bht

pmelange:
Modify the timing for a dfs reset at bht to 0-23 (request from Carsten)

Give the link to Segen a worse metric since the link quality bht-segen is so poor.  This makes the preference for internet access to go over perleberger36->strom-gw.

This image has already been built and is currently runing on bht-core


Noki (approved):
LGTM

PR #521, opened 2023-07-28
Wikiupdater: Made execution optional

noxilixon:
Added option so that the task is only executed when the wiki tag is given. Before it was executed every time.


FFHener (approved):
Looks good to me

Issue #265, opened 2022-06-28
Add ability to define extra "domain" entries in /etc/config/dhcp


pmelange:
This would be especially useful for a setup with uses unifi.  Having the unifi-controller off-site on a VM is very practical.  But in order to do this, the core router needs to resolv the hostname "unifi".  The needed section in /etc/config/dhcp would need to look like this:

```
config domain 
	option name 'unifi'
	option ip 'AA.BB.CC.DD'
```

spolack:
I'm currently not actively working on that.

PR #524, opened 2023-08-06
dhcp: add options for static leases and custom domain names

Noki:
This allows to add static dhcp leases and custom domain names by adding a `dhcp.yml` file to the location directory. This would resolve https://github.com/freifunk-berlin/bbb-configs/issues/265. The syntax of the dhcp.yml file is as follows:

```
---
dhcp_static_leases:
- name: Hostname1
  ip: 10.10.10.1
  mac: AA:BB:80:7C:E8:9E
- name: unifi
  ip: 10.10.10.2
  mac: AA:BB:80:7C:E8:99
dhcp_custom_domain_names:
- name: unifi
  ip: 10.10.10.2
```


Akira25 (approved):
Looks good to me at a short wentthrough.

One notice on the OLRS-services:
Hence the service name field has a flexible string, there are a lot of letters, which are not allowed to be there. Otherwise they will stop the olsrd from start up.

For an exact list of them, refer to the service-registrar sanitize code please.

As bbb-configs is an expert toop, it should be enough, to mention that circumstance in the documentation.

Akira25:
The aforementioned code is here:

https://github.com/freifunk-berlin/falter-packages/blob/master/luci/luci-app-falter-service-registrar/htdocs/luci-static/resources/view/falter-service-registrar/form.js

LL 9-11

Noki:
I added documentation. I would love to get some more feedback on this / active testing.

pmelange:
I know it's not in your code, but https://github.com/freifunk-berlin/bbb-configs/blob/b9cc657df7928f3079eb2527b109f93510aa1694/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2#L67 is causing problems.  It should be ```config domain```

Also, for my setup, I would prefer to limit the range of IPs on the dhcp network.  For example, I have it set up so that the start is 50 and the limit is 200.  I use the first 50 addresses for static leases and statically configured hosts (for prototyping and VM's).

pmelange:
These changes should also be done in templates/gateway/config/ too.

DEVELOPER.md
```diff
@@ -246,6 +246,55 @@ airos_dfs_reset:
...
+- name: webcam1
+  mac: CC:CC:CC:CC:CC:CC
+  ip: 10.10.10.3
+  dns: 1
+  olsrd_nameservice: 1
+- name: webcam2
+  mac: DD:DD:DD:DD:DD:DD
+  ip: 10.10.10.4
+  dns: 1
+  olsrd_nameservice: 1
```
pmelange, line 272: please remove trailing spaces

roles/cfg_openwrt/templates/corerouter/config/olsrd.j2
```diff
@@ -13,6 +13,16 @@ config LoadPlugin
 	list hosts '{{ network['prefix'] | ansible.utils.ipaddr(ip_num) | ansible.utils.ipaddr('address') }} {{ host }}'
   {% endfor %}
 {% endfor %}
+{% for dhcp_static_lease in dhcp_static_leases | default([]) %}
+{% if dhcp_static_lease['name'] is defined and dhcp_static_lease['ip'] is defined and dhcp_static_lease['olsrd_nameservice'] is defined and dhcp_static_lease['olsrd_nameservice'] == 1 %}
+        list hosts '{{ dhcp_static_lease['ip'] }} {{ dhcp_static_lease['name'] }}'
```
pmelange, line 18: It might be important to make sure that no hostnames are duplicated from the assignments list.  Those are added in lines 11-15

roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
```diff
@@ -75,3 +75,25 @@ config odhcpd 'odhcpd'
...
+
+{% endif %}
+{% endfor %}
+
+{% for dhcp_custom_domain_name in dhcp_custom_domain_names | default([]) %}
+{% if dhcp_custom_domain_name['name'] is defined and dhcp_custom_domain_name['ip'] is defined %}
+config domain
+        option name '{{ dhcp_custom_domain_name['name'] }}'
+        option ip '{{ dhcp_custom_domain_name['ip'] }}'
+
```
pmelange, line 97: First, L67 should be ```config domain``` but aside from that, it might be a good idea to check that hostnames are not repeated here from the assignments list.  And if a hostname is repeated, but with a different ip address (making sure to compare ipv4 with ipv4 and ipv6 with ipv6 addresses), then report it as an error.

DEVELOPER.md
```diff
@@ -246,6 +246,55 @@ airos_dfs_reset:
...
+
+```yml
+---
+olsrd_services:
+- name: Webcam 1
+  protocol: tcp
+  url: http://webcam1.olsr:80/
+- name: Webcam 2
+  protocol: tcp
+  url: http://webcam2.olsr:80/
```
pmelange, line 293: Please add two spaces to the beginning of each of these lines

DEVELOPER.md
```diff
@@ -246,6 +246,55 @@ airos_dfs_reset:
...
+  dns: 1
+  olsrd_nameservice: 1
+- name: webcam2
+  mac: DD:DD:DD:DD:DD:DD
+  ip: 10.10.10.4
+  dns: 1
+  olsrd_nameservice: 1  
+dhcp_custom_domain_names:
+- name: unifi
+  ip: 10.10.10.2
```
pmelange, line 275: Please add two spaces to the beginning of each of these lines

pmelange (changes requested):

pmelange:
> These changes should also be done in templates/gateway/config/ too.

I still don't see any changes in templates/gateway/config/

Noki:
Yes, not done yet... added all open tasks to the todo list of this PR (first post).

Noki:
Der aktuelle Stand ist, dass es eigentlich erst sinnvoll ist dies fertig zu machen, wenn wir alle Geräte als Hosts beschreiben, also auch Antennen und Switches und die SNMP Profiles, DFS-Reset und eben auch services und leases dann auf die Hostbeschreibung wechseln. - So war zumindest mal der Plan den ich mit @FFHener besprochen hatte.

PR #526, opened 2023-08-10
l105-gw: add eth1.32 interface to mesh with b.bbb-vpn

pmelange:
This interface was lost while being migrated to the bbb-configs.  Now it's back.  This system is already running on the vmhost and is routing traffic.


Noki:
Looks good to me

PR #527, opened 2023-08-15
templates: check 15 character limit for bridges

### LINKED ISSUE #185: device name length can exceed 15 chars
Bridge-interfaces are are prefixed with "br-", so device can exceed maximum lengths even if they are just 15 chars.

Noki:
This commit adds the 15 character limit check for bridge interfaces and adjusts the condition to actually check for 15 characters.  This resolves #185. The 15 character limit is mentioned here: https://openwrt.org/docs/guide-user/network/tunneling_interface_protocols


robertfoss:
LGTM

robertfoss (approved):

PR #529, opened 2023-08-18
ak36: update IP and interface comments

pktpls:
Add comments for IPs of new ak36-teufelsberg and ak36-philmel60 devices as well as the strange hypervisor IP.

The respective mesh interface are commented out because the corerouter/gateway VM doesn't have them yet.


Noki:
looks good to me

PR #536, opened 2023-08-30
dnsmasq: Disable rebind protection

robertfoss:
Currently RFC1918 addresses are not resolving due to the rebind attack protection. This is a problem if you for example publish internal VPN IP addresses on a public DNS server.

Affected ranges:
10.0.0.0/8
192.168.0.0/16
172.16.0.0/12


Noki:
Can you please describe why you want to deactivate the protection? What is your use case?

I don't think it is a good idea to disable this for all DHCP networks, maybe on private DHCP networks not associated with berlin.freifunk.net.

pktpls:
The use case seems to be internal IP addresses being resolved from public DNS records. You can make the exemption less broad with `--rebind-domain-ok=[<domain>]|[[/<domain>/[<domain>/]`

Generally for such cases Split DNS is perfect:

```
server=/corp/1.2.3.4
```

Requests for records within the `.corp` TLD will be forwarded to DNS server 1.2.3.4. I do something similar at home (but on Fedora/NetworkManager/systemd-resolved) to send only `.olsr` requests to the Freifunk router.

BUT of course probably this is out of your control and customizing rebind protection might still be neccessary.

robertfoss:
@Noki My usecase is just as @pktpls describes it.

Manually configuring Split DNS for every machine connected to the network is unmaintainable.

The security benefits of this feature are negligible I would say. Rebind protection isn't commonly deployed as far as I can tell. 

If you still prefer it to be the default shipped, I could make it configurable on a per node basis.

Noki:
> The security benefits of this feature are negligible I would say. Rebind protection isn't commonly deployed as far as I can tell.

DNS rebind protection is on by default on most consumer routers e.g. AVM Fritzbox, Telekom Speedport so I think it is commonly deployed.

> If you still prefer it to be the default shipped, I could make it configurable on a per node basis.

Yes, please make it configurable per node or if possible per VLAN so it can be used in a private network setting, but still be on for VLAN 40.

pktpls:
It just occured to me that rebind protection is about protecting the private LAN, and we maybe don't really need that because the private LAN is inaccessible in bbb-configs - it's isolated in the uplink namespace.

I'm okay with disabling it. Would also allow people to bind public domain names to their OLSR services.

robertfoss:
@Noki Are you ok with @pktpls conclusion?

Noki:
@robertfoss I do disagree with @pktpls conclusion. Computers can be connected to the Freifunk network and and other private networks at the same time and from what I understand disabling the DNS rebind protection on the Freifunk network can open up the attack vector on the private network.

If you want to completely disable the rebind protection, you should make it a setting. Another possibility would be to specify an exception using the [rebind_domain](https://openwrt.org/docs/guide-user/dhcp/dhcp_configuration#:~:text=protection%20is%20enabled-,rebind_domain,-list%20of%20domain) which would be even better in my opinion.

Note: adding exceptions is also the way to bypass dns rebind protection in Fritz!Box routers.

robertfoss:
I think I'll just close this ticket. I don't have bandwidth to dig into this at the moment.

PR #547, opened 2023-10-03
model: switch mt7915 devices to 23.05

Noki:
Mesh is stable now in 23.05 with the latest fixes, therefore switch the devices over to 23.05


pktpls:
I also extensively tested 23.05 on the NWA50AX the last few days :+1:

PR #548, opened 2023-10-03
all: update dns servers and add ipv6

pktpls:
- Digitalcourage updated their recommended addresses
- IPv6 for both Digitalcourage and FDN
- bug: AS250 anycast unreachable for us at L105


pktpls:
The thing at L105 is that AS250.net have a presence there as well, while we actually route everything from L105 via AK36. All other locations "correctly" route AS250.net to Frankfurt. Either AS250.net's anycast at L105 is broken, or our route-via-AK36 setup at L105 has an issue.

Noki (approved):
looks good to me

PR #549, opened 2023-10-03
networks: add tunnel role for use with tunspace

pktpls:
TunSpace is a new attempt at a client for wireguard tunnels - see freifunk-berlin/falter-packages#413 for the source code of TunSpace itself. The server side is still the same setup using wg-installer.

The preliminary syntax for this is:
```yml
role: corerouter
openwrt_version: 23.05-SNAPSHOT  # or snapshot
host__packages__to_merge:
  - -tunnelmanager
  - tunspace
  - wireguard-tools

networks:
  # ... other stuff first ...
  
  - vid: 50
    role: uplink

  - role: tunnel
    ifname: ts_wg0
    mtu: 1412
    prefix: 10.31.123.42/32  # ipv6 prefix is always fe80::2/64
    wireguard_port: 51820
    mesh_metric_lqm:  # optional
      - 'default 0.5'
    mesh_metric: 256  # optional

  - role: tunnel
    ifname: ts_wg1
    # ...
```

Please try it out, for me it has been working well for the past few days. Note that it currently logs to syslog *a lot* - this is intended, we'll make it quieter soon when we're confident it works. Until then you can set `option debug 0` in /etc/config/tunspace.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -155,28 +155,29 @@ The VLAN ID (vid) usually follow this numbering convention.
 ##### uplink via tunnel
 
 If for some reason you are in need of an uplink via a "normal" internet connection, a wireguard
-tunnel can be an easy and safe solution. In that case add another vlan to the networks.yml
+tunnel can be an easy and safe solution. In that case you can add one or more tunnel networks:
 
 ```yml
-  - vid: 50
-    untagged: true            # option to don't tag this vlan - useful if the corerouter is plugged into a normal home router
-    name: uplink
-    role: uplink
-    tunnel_wan_ip: 192.168.1.2/24   # put here the address and subnet of the corerouter inside the uplink network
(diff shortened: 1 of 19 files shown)


PolynomialDivision:
Awesome! Thanks! :) Running that in my network and works fine so far.

pktpls:
I will change three more little things:
- The nftables MSS rules need to be added in templates/corerouter/nftables.d/20-wg-maxseg-size.nft.j2
- The tunnel MTU needs to be smaller in some cases, e.g. PPPoE or DS-Lite. The default should be 1280 which is always safe. You can go higher on specific locations where you know the exact uplink MTU.
- Yesterday I witnessed how sometimes during boot, tunspace might create the uplink namespace and macvlan interface before the actual uplink interface is ready. Under certain circumstances this fails and we only get uplink DHCP on the next maintenance interval.

pktpls:
@PolynomialDivision said "let's merge it" and I updated developer.md

PR #555, opened 2023-10-09
ansible: move away from limited /tmp ramdisk, and clean up dir structure

pktpls:
Working in /tmp is great because it's a ramdisk (fast & ephemeral), but it's also bad because it's a ramdisk (space-constrained & ephemeral). We can only build a few hosts at the same time, at a certain number we run out of diskspace (ramspace). That makes it hard to mass-test image builds locally, and it can also fail github builds, e.g. https://github.com/freifunk-berlin/bbb-configs/actions/runs/6445019125/job/17498618384

Instead we'll now build in ./tmp in the bbb-configs directory. While we're at it, clean up the various directory path variables.

### DIFF
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -46,4 +46,4 @@ jobs:
         uses: actions/upload-artifact@v1
         with:
           name: ansibleimages
-          path: /tmp/ansible-openwrt/images
+          path: ./tmp/images
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .vscode/
 venv/
 .env
+tmp/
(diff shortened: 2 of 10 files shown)


Akira25:
I understand the need for changing this, when mass flashing, but most of the users do flash only one site at once. Maybe we can have this configurable somehow? Then we can have the advantages of both approaches.

pktpls:
I don't feel like working in `/tmp` has any big advantages, I tried ramdisks with buildbot and even on those slow VMs, there wasn't really any performance advantage.

Is anyone's primary computer harddisk-based? :)

Noki:
looks good to me. please fix the linter issue.

pktpls:
```
/github/workspace/FAQ.md:89:401 MD013/line-length Line length [Expected: 400; Actual: 1018]
```

I didnt touch that :-(

PR #556, opened 2023-10-09
all: add hener to list of trusted ssh keys

Noki:
In my opinion, hener has proven that he is a committed and responsible member of our community. I suggest therefore we add his SSH key to the list of often trusted keys of all mayor locations.


Akira25 (approved):
Looks good to me. Good decision, in my opinion.

booo (approved):

pmelange (approved):

PR #572, opened 2023-11-13
strom: removed non existing antennas

Noki:


PolynomialDivision:
I think the antennas exist but did not survive flashing. ^^

PR #577, opened 2023-11-15
teufelsberg: add rocket-sectors

Koltonowski:


Noki:
please fix the linter issues!

pktpls:
It's fine, my editor will remove that one trailing space character next time.

We really need to discuss those style linters. They don't fix a single concrete problem. I am even in favor of consistent coding style, but they way we do it right now is counterproductive and very frustrating.

pktpls:
(Shellcheck is useful, and other tools that actually check the semantic meaning of code)

Noki:
I just corrected it. Of course we can discuss linter rules and disable some, however in yaml identation has a meaning and should therefore be checked.

PR #580, opened 2023-11-19
rio: adjust routing towards emma

Noki:
the link towards emma is poor, therefore adjust mesh metrics so traffic from simeon flows via rhxb, rhnk towards emma


FFHener (approved):
Lgtm, it might make sense to try to realign the antenna after the upgrades at emma and have a look again at these values

PR #583, opened 2023-11-21
bringup ska95

spolack:


locations/ska95.yml
```diff
@@ -48,6 +53,11 @@ networks:
     assignments:
       ska95-core: 1
 
+  - vid: 41 # Private LAN where private Internet is connected
+    role: ext
```
pktpls, line 58: This is called `uplink` now

Noki:
The geolocation is wrong. Please have a look at hopglass.

pktpls:
And I think some IPs are mixed up, or is it indeed 10.31.8.x *and* 10.36.8.x?

locations/ska95.yml
```diff
@@ -0,0 +1,77 @@
+---
+
+location: ska95
+location_nice: Skaliter Str. 78
+latitude: 52.52075
+longitude: 13.40948
```
Noki, line 6: should be 52.500043, 13.43242

locations/ska95.yml
spolack: But then i'd question the name. why should it be uplink when there is no uplink?

locations/ska95.yml
spolack: thx!

locations/ska95.yml
pktpls: I always thought the purpose of the `ext` role was to attach an uplink network, usually DSL or mobile. I've never seen it used without a tunnel

spolack: i was probably the only one who ever has used it. :)

pktpls: It's a legit use case, it's just news to me :-)

Noki:
I fixed the typo in the street name and adjusted the location based on flb2net's input.

locations/ska95.yml
Noki: ok, so the ext changes / fixes are fine? can we merge this?

PR #589, opened 2023-11-24
mela: init location

Noki:


FFHener (changes requested):
It might make sense to put the OpenWRT-Devices into different frequencies, althoutgh it shouldnt make to much of a difference right now when only 2 devies are running

Noki:
> It might make sense to put the OpenWRT-Devices into different frequencies, althoutgh it shouldnt make to much of a difference right now when only 2 devies are running

In general I agree, however I think it is better to start with the default channels, so people can mesh with a standard falter image. We can always change the frequencies later if there is some usage.

FFHener (approved):
Lgtm

PR #594, opened 2023-11-25
models: some 23.05 updates

pktpls:


PolynomialDivision:
maybe remove all unused models?

pktpls:
I was thinking the same. Only two currently unused models will definitely be used in the future (netgear_wax202 & zyxel_nwa55axe)

Noki:
CPE501 will also be used as there are some on mela

PR #599, opened 2023-11-26
network: support untagged VLANs on devices without a switch

### LINKED ISSUE #598: untagging a VLAN does not work on devices without a switch
on a device without a switch I end up with

```
config device
        option name 'br-uplink'
        option type 'bridge'
        list ports 'lan.50'
```
with a VLAN 50 set to `untagged: true` as https://github.com/freifunk-berlin/bbb-configs/blob/db01dee840d08e6f4181c23a5162a9d85a57f20f/roles/cfg_openwrt/templates/common/config/network.j2#L23 does not handle the setting.

Noki:
This fixes #598 and enables untagged VLANs on devices without a switch, e.g. DAP-X1860 so the network port can be used to connect to an untagged network in an uplink configuration or to untag the DHCP network.


robertfoss:
lgtm

PR #609, opened 2023-11-29
location_philmel: convert to new location file and add new accesspoints

Petrella:
Philmel needed some adjustements due to adding some access points with special config and ssid names


group_vars/all/wireless_profiles.yml
Noki, line 1: This file should not be changed. Wireless_profiles moved to the locations and should not be added back here.

Noki (changes requested):
When switching to the single-file-format, the definitions needs to be moved. Please delete the files in group_vars and host_vars so they only exist in locations/philmel.yml from now on.

FFHener:
Please mark the pr as a draft until its ready for a review

FFHener:
Closing as there is a new version of this PR

PR #616, opened 2023-12-05
philmel: convert to single file format

Noki:


Noki:
I just set this to "ready for review". Philmel requires additional changes from #609 but this can be done afterwards.

FFHener (approved):
All the things that should be changed seemed to be there before, so i think we can merge it for the moment

PR #617, opened 2023-12-05
zwingli: updated and converted to single file location

FFHener:


Noki:
It would really help to have this in 2 commits so reviewing is easier: 
* migration to single file location
* changes

I noticed zwingli-nno-nf-5ghz misses a mesh VLAN. This should be added or it should be clarified why we don't have it.

In addition I think we should comment out the defect devices in assignments, so we will have a new PR when the replacement happens.

FFHener:
> It would really help to have this in 2 commits so reviewing is easier:

Thanks for pointing out, i will do so in the next conversion to SFL

> I noticed zwingli-nno-nf-5ghz misses a mesh VLAN. This should be added or it should be clarified why we don't have it.

Its likely due to another antenna pointing north already having mesh activated and the nno-nf-5ghz Antenna doesnt use the default Frequency for meshing - so i think this was intentionell by the people who builded up this location. I added a comment regarding this topic.

> In addition I think we should comment out the defect devices in assignments, so we will have a new PR when the replacement happens.

Done

PR #619, opened 2023-12-10
Q216 update

FFHener:
Converted to single file location, removed outdated wireguardtunnel and added new ptp-Antenna


locations/q216.yml
```diff
@@ -15,6 +48,12 @@ networks:
     prefix: 10.31.201.249/32
     ipv6_subprefix: -10
 
+  - vid: 10
```
Noki, line 51: this vlan is already defined. should be 11

PR #620, opened 2023-12-10
Several conversions to Single File Format

FFHener:
I will add more locations during the next days.
Not to shure thought what to do with `group_vars/location_habersaath/NETWORKS.md`


FFHener:
Ansible Test Build fails due to diskspace limitations

FFHener:
Will switch to individual PRs to make the review process better to handle

PR #623, opened 2023-12-10
fanninger47: init

FFHener:
This location will be build up in the near future


PolynomialDivision:
no mac override?

FFHener:
Added mac override. It should be added in the documentation why its needed

PR #626, opened 2023-12-11
agym: converted to /locations

FFHener:


locations/agym.yml
```diff
@@ -1,4 +1,27 @@
 ---
+
+location: agym
+location_nice: Andreas Gymnasium
+latitude: 52.514169
+longitude: 13.433311
+community: true
+
+hosts:
+
```
PolynomialDivision, line 10: why the extra new line?

FFHener: I liked it more with more spacing. I thought it might help with clarity. But I dont have a strong opinion about it. If you want to i can change it

PR #627, opened 2023-12-11
Casa kua: converted to /locations and renamed to use `-` instead of `_`

FFHener:


FFHener:
Test build fails due to the renaming. It uses a wrong `--limit` in the Ansible Test Build. Testbuild runs fine locally

locations/casa-kua.yml
```diff
@@ -1,30 +1,30 @@
...
+  - hostname: casa-kua-core
     role: corerouter
     model: "glinet_gl-b1300"
     host__packages__to_merge:
-          - -tunnelmanager
-          - tunspace
-          - wireguard-tools
+      - -tunnelmanager
+      - tunspace
+      - wireguard-tools
```
PolynomialDivision, line 17: we don't need that anymore

PR #638, opened 2023-12-13
DEVELOPER.md: Rewrite for locations/

FFHener:
Rewrite of the file to account for the new standard of using single file locations. Added wireless profiles and some other smaller hints as well. It would make sense to write better docs for the profiles in the future.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -4,11 +4,270 @@ This guide explains the directory structure of the project and variables within 
 
 ## locations/
 
-`locations/` holds a single file format for locations as an alternative to the old way of using `groups_vars/` and `host_vars/`. Instead everything is simply written into a single file named `$LOCATIONNAME.yml`.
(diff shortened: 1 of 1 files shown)


FFHener:
Testbuild failed due to an connection timeout to openwrt. It can be ignored as no components regarding ansible has been touched.

```
TASK [cfg_openwrt : Download Imagebuilder if upstream is newer] ****************
fatal: [l105-gw]: FAILED! => {
    "changed": false,
    "dest": "/home/runner/work/bbb-configs/bbb-configs/tmp/dl//openwrt-imagebuilder-23.05-SNAPSHOT-x86-64.Linux-x86_64.tar.xz",
    "elapsed": 10,
    "url": "https://downloads.cdn.openwrt.org/releases/23.05-SNAPSHOT/targets/x86/64/openwrt-imagebuilder-23.05-SNAPSHOT-x86-64.Linux-x86_64.tar.xz"
}

MSG:

Connection failure: The read operation timed out
```

PR #641, opened 2023-12-13
Strom: converted to locations/ and clean up

FFHener:


locations/strom.yml
```diff
@@ -1,29 +1,42 @@
...
-
 uplink:
   ifname: eth0.1301
   ipv4: 77.87.51.131/25
   # ipv6: none
 
 mgmt:
   ifname: eth0.42
-  ipv4: 10.31.130.65/27
+  ipv4: 10.31.130.64/27
```
PolynomialDivision, line 31: this is the network address? Are you sure that this is correct?

locations/strom.yml
```diff
@@ -1,29 +1,42 @@
...
+longitude: 13.342569613915
+community: true
+
+hosts:
+
+  - hostname: strom-gw
+    role: gateway
+    model: "tplink_tl-wdr4900-v1"
+
+ipv6_prefix: "2001:bf7:750:2400::/56"
```
PolynomialDivision, line 15: Why did the ipv6 prefix change?

locations/strom.yml
```diff
@@ -1,29 +1,42 @@
...
+location_nice: Stromstraße 2A
+latitude: 52.523692495967
+longitude: 13.342569613915
+community: true
+
+hosts:
+
+  - hostname: strom-gw
+    role: gateway
+    model: "tplink_tl-wdr4900-v1"
```
PolynomialDivision, line 13: please, lets finally exchange this router xD

locations/strom.yml
FFHener: The `.65` is the first usable hostadress, but from my understanding we are defining this ipv4-prefix with a network address, so technically `10.31.130.64/27` would be right, although it shouldn't change how ansible behave.

```
> sipcalc 10.31.130.65/27
-[ipv4 : 10.31.130.65/27] - 0

[CIDR]
Host address		- 10.31.130.65
Host address (decimal)	- 169837121
Host address (hex)	- A1F8241
Network address		- 10.31.130.64
Network mask		- 255.255.255.224
Network mask (bits)	- 27
Network mask (hex)	- FFFFFFE0
Broadcast address	- 10.31.130.95
Cisco wildcard		- 0.0.0.31
Addresses in network	- 32
Network range		- 10.31.130.64 - 10.31.130.95
Usable range		- 10.31.130.65 - 10.31.130.94

-
```

pktpls: .65 was correct, it's the first usable address in the network named .64, see "Usable range" in sipcalc's output

PR #645, opened 2023-12-14
rigaer78: convert to single file format

freifunkoperator:


host_vars/rigaer78-back-floor-2-kitchen/base.yml
```diff
@@ -1,9 +0,0 @@
----
-
-location: rigaer78
-role: ap
-model: "avm_fritzbox-4040"
-
-# only useful in combination with port_untag patch that is not upstream
```
FFHener, line 7: Is this comment no longer valid?

locations/rigaer78.yml
```diff
@@ -1,4 +1,97 @@
...
+  - hostname: rigaer78-zwingli
+    address: 10.31.119.132
+    snmp_profile: airos_8
+
+hosts:
+  - hostname: rigaer78-core
+    role: corerouter
+    model: "bananapi_bpi-r64"
+    wireless_profile: freifunk_default
+    dhcp_no_ping: false
```
FFHener, line 22: Why is this set to false as mgmt only contains static adresses?

Noki:
In addition 3 devices are in outage or do not exist anymore:

rigaer78-back-floor-0-garage
rigaer78-back-floor-3-left
rigaer78-eap225

pktpls:
LGTM, cleanups can happen another time

PR #651, opened 2023-12-14
bht: converted to locations/

FFHener:


Noki:
bht-jup is offline. What is the state of that antenna is it still there and needs to be replaced or should we remove it from the config?

FFHener:
I asked and will update the PR accordingly

pmelange:
That dish should still be there, but somehow is not working.  Keep the config and the next time someone goes up there, they can try to get it back online.

pktpls:
jup-bht has a wifi connection to the `freifunk-bbb-beuth-o` SSID, not sure why bht-jup isn't reachable

Noki:
as pmelange pointed out the bht-jup antenna is in outage and needs to be checked / replaced at some point. just merge this for now, we also have the outage in UISP.

PR #654, opened 2023-12-14
Forcki: converted to locations/ and better `location_nice`

FFHener:


Noki:
This location is offline. We should check if something needs to be fixed or if the location can be removed.

FFHener:
I pm'ed the person who builded up this location.

FFHener:
The current status of the location is that it was builded up and "sometimes works" according to one maintainer, so it would merge this PR for now.

PR #656, opened 2023-12-14
koepi: converted to locations/

FFHener:
We might as well just delete this location as it is no longer existing as far as i can tell


FFHener:
Got out of Band Approval. It is plannend to build it up again in the near future

PR #657, opened 2023-12-14
liese-21: converted to locations/ and removed old ssh-key

FFHener:


FFHener:
Testbuild works locally. It fails because the folder used to be named `location_liese_21` instead of using a `-`.

locations/liese-21.yml
```diff
@@ -1,5 +1,46 @@
...
+    # We use Ubiquiti 802.11af-to-passive converters on all ports.
+    # They neogotiate more power than they actually draw,
+    # so we need to increase poemgr's power budget.
+    # Without this increase, devices will be not powered up randomly.
+    poemgr_power_budget: 24
+
+  - hostname: liese-21-n-nf-5ghz
+    role: ap
+    model: "mikrotik_sxtsq-5-ac"
+    # TODO: Add mac_override
```
Noki, line 25: please add this

PR #659, opened 2023-12-14
mlk-nk: converted to locations/

FFHener:


FFHener:
Build doesnt fail locally with `ansible-playbook play.yml --limit mlk-nk-'*'`

PR #663, opened 2023-12-15
rhnk: added rhnk-nno-60ghz, minor fixes

Noki:
Added the new Wave AP towards Zwingli
Added SNMP monitoring, which should work with AF60 profile now


FFHener (approved):
Looks good to me. If snmp-Monitoring works we should roll it out for the other Wave Devices

PR #665, opened 2023-12-15
rhnk: remove devices

Noki:


locations/rhnk.yml
```diff
@@ -97,13 +85,6 @@ networks:
     mesh_metric: 128
     ptp: true
 
-  - vid: 13
```
FFHener, line 100: Richardplatz is connected to rhnk-ssw-5ghz, so thanks for cleaning up

PR #666, opened 2023-12-15
schleidenplatz: converted to locations/

FFHener:


Noki:
Looks offline to me. Is this still an active location or should we remove it?

FFHener:
I didnt build up this location, so i have no clue about the current state. Just converted it. We can remove it if its no longer exist.

spolack:
No idea tbh. Lets merge for now.

PR #670, opened 2023-12-16
Faster build times & little inventory fixes

### LINKED ISSUE #661: Ansible Test Build: Better handling of locations using dashes
The way ansible and everything works the testbuild currenly fails because of ansible-underscore-foo. If we switch it to use a wildcard this problem is solved.
[Example mlk-nk](https://github.com/freifunk-berlin/bbb-configs/actions/runs/7214128983/job/19655535111)

pktpls:
- Make inventory a little bit faster (= half as slow). Singlefile locations are slow to parse because it's a new python interpreter for every single host. This gets worse the more locations we convert to singlefile format. `yj` is a golang tool which has much faster startup. I'm not happy adding a non-python dependency, but I want to make this faster right now make it "pretty" later. Ideally we'd rewrite the inventory script in python - not starting any new interpreters means no need for a golang tool.
- only build if actual code has changed, not for any non-code changes
- only build configs, not full images
  - this has been very time consuming and error prone when trying to merge PRs. we all have limited time and shouldn't spend it waiting for slow builds with limited value and reliability.
- Fix change detection for github worker testbuilds. The singlefile location filenames need to be turned into `location_foo_bar` tags. (fixes #661)
- fix test build AP hostname
- ~~Fix parzelle17-fritz location value.~~ (#690)

### DIFF
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -30,7 +30,7 @@ jobs:
           echo "Building for $CHANGED_HOSTS"
           echo "CHANGED_HOSTS=$CHANGED_HOSTS" >> "$GITHUB_ENV"
           CHANGED_LOCATIONS=$(git diff --diff-filter=d --name-only "origin/$BRANCH" | grep 'group_vars/location_' | sed 's/group_vars\/\(.*\)\/.*/\1/' | sort | uniq | sed -z 's/\n/,/g;s/,$//')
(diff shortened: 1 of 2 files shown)


pktpls:
> Fix change detection for github worker testbuilds

Mh this won't work...

pktpls:
Ok this is fine now - 34s testbuild time because no "actual" code changed.

FFHener:
For me it looks good. I think only building the config is a trade off i can live with as imo we have more false positives than actually catching errors. And we will build a host anyway again if we deploy the config

FFHener (approved):
Lgtm

PR #672, opened 2023-12-16
rhnk: fix nf APs and mesh

Noki:
This fixes the nf APs. Without this the EAP225 uses too much power and does not account for the additional antenna gain. In addition this is required so mlk-nk can mesh again via 802.11s.


FFHener (approved):
Looks good to me

PR #673, opened 2023-12-16
mlk-nk: multiple changes and fixes

Noki:
- config now matches the state of the location
- VLANs now match our convention
- corrected settings of so-antenna
- renamed antenna from zwingli to nno and made it an AP
- removed incorrect channel settings

### DIFF
--- a/locations/mlk-nk.yml
+++ b/locations/mlk-nk.yml
@@ -13,10 +13,6 @@ hosts:
     model: "tplink_tl-wdr4900-v1"
     wireless_profile: freifunk_default
 
-  - hostname: mlk-nk-nf-wbp1
-    role: ap
-    model: "ubnt_nanostation-m2_xm"
-
   - hostname: mlk-nk-nf-wbp2
     role: ap
     model: "ubnt_nanostation-m2_xm"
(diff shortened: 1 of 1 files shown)


locations/mlk-nk.yml
```diff
@@ -129,22 +117,11 @@ networks:
...
 
-location__channel_assignments_11a_standard__to_merge:
-  # Channel - Bandwidth - TX Power
-  mlk-nk-rhnk: 36-20
-
-location__channel_assignments_11g_standard__to_merge:
-  # Channel - Bandwidth - TX Power
-  mlk-nk-nf-wbp1: 1-20
-  mlk-nk-nf-wbp2: 6-20
-  mlk-nk-nf-wbp3: 11-20
```
FFHener, line 146: Why did you removed the channel assignments for 2,4GHz?

FFHener (commented):
Besides the channel assignments everything looks good to me

locations/mlk-nk.yml
Noki: Because I checked the devices and they were set to the default (13-20). The configuration was not matching the setup here.

PR #675, opened 2023-12-18
zwingli: reactivate ap with new hardware and activate wifi_roaming

noxilixon:


noxilixon:
The zwingli-dom-1 ap was flashed and installed, but the cable is not working correctly. So this device is still not running. This config should represent the current state of the location and will be merged as soon as the device is working again.

PR #684, opened 2023-12-19
Include wikiupdater into testbuild action

FFHener:
Make the wikiupdater skript automaticly updates wikiarticles when a push happens to `master`. This will help keeping our wiki always up to date. I added the option `continue-on-error: true` so we will go on if no wikiarticle was found. In addition I created credentials for a `bbbot`. They are stored inside this repo in the envirement called [wikiupdater](https://github.com/freifunk-berlin/bbb-configs/settings/environments/1751983185/edit)

### DIFF
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -1,6 +1,12 @@
 ---
 name: Test Build
-on: [pull_request]  # yamllint disable-line rule:truthy
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/testbuild.yml
```diff
@@ -1,6 +1,12 @@
 ---
 name: Test Build
-on: [pull_request]  # yamllint disable-line rule:truthy
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
Noki, line 9: I think this should not run when performing a pull request, only with a push to master.

FFHener: The testbuild is supposed to run on pull requests. This is the behavior from before. I added if conditions starting from [Line 47](https://github.com/freifunk-berlin/bbb-configs/pull/684/files#diff-fb5c483c388bb8411369b8ae18c01c14b2389999d00b8f6cdd898a27d8282cd1R47) that differentiate between PR and push

Noki: sorry, missed that one...

.github/workflows/testbuild.yml
```diff
@@ -38,11 +44,19 @@ jobs:
           echo "Building for $CHANGED_LOCATIONS"
           echo "CHANGED_LOCATIONS=$CHANGED_LOCATIONS" >> "$GITHUB_ENV"
 
-      - name: Build changed locations
+      - if: ${{ github.event_name == 'pull_request' }}
```
Noki, line 47: this should still happen on pull_request and push as we still have this one:

image

FFHener: The original code states
```
name: Test Build
on: [pull_request]
```
so it was running only on PRs. But as we discussed this bypass before i would just go and remove it.

FFHener:
Converting it to a draft as there are currently some wiki articles that would get wrongly modified

PR #687, opened 2023-12-21
model: Mikrotik Routerboard 750gr3 bump OpenWRT version

Noki:
this currently affects sgfrd-core and vaterhaus-core. We should flash this on sgfrd-core first, since it's easier to access if something goes wrong, but this is a mt7621 target, so I don't expect any issues.

@Petrella please test this on sgfrd-core.


Petrella (approved):
Looks nice to me. Though not testet yet on real device.

PR #689, opened 2023-12-21
Wikiupdater: more robust way of getting wiki articles

FFHener:
As the old way of getting an wikiarticle by a node name used by `util.berlin.freifunk.net` was unrelieable as it was linking all nodes in a certain area to a single article we now use a fixed relation. Every location has an article at `Berlin:Standorte:$LOCATION`. If the actual article is located somewhere else this page should get a redirect to the correct article. The python script will follow the redirect.

ska95: Updated the role to `uplink` as this is the new name used by tunspace.

once per location: As we now use `$LOCATION` instead of `$HOST` we can archive running the script only once per location in this new way. In addition it is now possible to update an wikiarticle even if the config run doesnt include the core-router.

This should be merged before #684


FFHener:
Redirects done until `noki-core`

locations/ska95.yml
```diff
@@ -55,7 +55,7 @@ networks:
       ska95-core: 1
 
   - vid: 41  # Private LAN where private Internet is connected
-    role: ext
+    role: uplink
```
Noki, line 58: In this case it's really simply an external private network that is connected to the core so the antennas could be managed from there. No uplink is done via that network. This should not be changed. There was a discussion about this in the emma chat. So we still have the ext role, but for uplinks we use uplink now.

Noki (approved):
looks good to me

PR #690, opened 2023-12-22
parzelle17-fritz: fixed location name

FFHener:
Testbuild fails due to handling of `-`, #670 would fix it


Noki (approved):
looks good to me

PR #691, opened 2023-12-24
gateways: added assignments for olsrd nameservice

Noki:


FFHener:
Linter fix commit could be squahed, right? Besides that i didnt feel comfortable merging it myself right now as i dont know the reson why they used comments in the first place.

spolack:
LGTM. Comments were used because it was not implemented back then.

PR #692, opened 2023-12-28
model file: add glinet gl mt1300

Petrella:
Enable image and config generation for the glinet gl mt1300 travel router


pktpls:
Herzlich willkommen

PR #699, opened 2024-01-07
inventory: back to acceptable performance

pktpls:
We have ~300 hosts in ~100 locations and we were calling `yq` multiple times for every single one of them. That was ~700 python interpreter invocations.

Now we only call it once to convert all YAML to JSON in one go, and let one `jq` script do all the merging. It's less code and much faster.

- Before: `ansible-inventory --list` 0m35,404s
- After: `ansible-inventory --list`  0m4,649s
- Without Ansible: `./inventory/base_inventory --list`  0m0,543s

### DIFF
--- a/inventory/base_inventory
+++ b/inventory/base_inventory
@@ -1,134 +1,39 @@
 #!/usr/bin/env bash
 
-# shellcheck disable=SC2086
-# shellcheck disable=SC2002
-#
-# SC2086: Double quote to prevent globbing and word splitting.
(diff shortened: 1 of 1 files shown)


pktpls:
Habe noch Verbesserungen für inventory/README.md rumliegen, aber das dauert noch ein bisschen und der Performancegewinn hier ist so groß dass ich es schon mal vorschiebe.

Noki (approved):
looks good to me

PR #700, opened 2024-01-07
gateway: allow inbound for cryptpad.berlin service

pktpls:
I'm setting up a Cryptpad instance :lock:


Noki:
I think we should start adding information about which core-router handles the prefix and also state this as mandatory.

pktpls:
> which core-router handles the prefix

The cryptpad host won't be in bbb-configs

Noki (approved):
looks good to me

pktpls:
Hm, this should have built the configs for all gateway hosts, but didn't. I'll fix that separately.

PR #702, opened 2024-01-07
flughafen: use 5ghz legacy station antenna as south AP

Noki:
This antenna was unused and in station mode, therefore reconfigure it to be a south facing ap for people to connect.


FFHener (approved):
lgtm

PR #706, opened 2024-01-09
wilde: add EAP225-Outdoor, disable mesh on APs

pktpls:


FFHener (approved):
lgtm, but we could consider adding this wireless profile as a general one in the future.

Noki:
yes, please do as proposed by @FFHener as there are some other locations that would benefit from this

PR #707, opened 2024-01-10
dragonkiez: 23.05 and single file format

Noki:
This PR switches the unifi ac mesh over to 23.05 as builds are broken with 22.03.
This PR converts dragonkiez to single file format.

Note: As we discovered just flashing unifi ac mesh devices remotely with this will brick the devices, as partitions are enlarged between OpenWRT releases. The devices therefore need to be updated manually at the location. The tested migration path is:

22.3-Freifunk -> 22.3-OpenWRT -> 23.5 OpenWRT -> 23.5-Freifunk

### DIFF
--- a/group_vars/location_dragonkiez_adlerhalle/general.yml
+++ b/group_vars/location_dragonkiez_adlerhalle/general.yml
@@ -1,2 +1,0 @@
----
-community: true
--- a/group_vars/location_dragonkiez_adlerhalle/owm.yml
+++ b/group_vars/location_dragonkiez_adlerhalle/owm.yml
@@ -1,4 +1,0 @@
----
-location_nice: Dragonkiez Adlerhalle
-latitude: 52.495065429
-longitude: 13.387666941
--- a/group_vars/location_dragonkiez_adlerhalle/snmp.yml
+++ b/group_vars/location_dragonkiez_adlerhalle/snmp.yml
@@ -1,6 +1,0 @@
----
-
-snmp_devices:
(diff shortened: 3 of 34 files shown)


pktpls:
Did you test it with a device? Broken builds are acceptable imho

pktpls:
What I mean is, I think after yesterday I prefer tested upgrades over green builds

Noki:
I did not test it, therefore it is still a draft. ;-)

Noki:
@pmelange tested this and with this we will brick the devices. This needs additional testing...

Noki:
In my opinion, this PR is ready for review and merge because we have shown in tests that the devices can run with 23.05 based BBB configs images. The only important thing to keep in mind here is that the upgrade path requires an on-site appointment.

pktpls:
> partitions are enlarged between OpenWRT releases

Ah yes that makes sense - our 22.03-based build failed because the image turned out too large (7461470 bytes). The SquashFS padding steps turn this into 9764864 bytes, but the `kernel0` partition is only 7929856 bytes.

```
...
7461470 bytes (7.5 MB, 7.1 MiB) copied, 0.0792106 s, 94.2 MB/s
padding image to 00950000
bash: line 1: 8: Bad file descriptor
...
```

They did a similar resize by combining `kernel0` and `kernel1` partitions with the Unifi Switch Flex and others.

pktpls:
LGTM :+1: 

Btw the same procedure applies to ubnt_unifiac-pro @ hds-ffraum.

PR #708, opened 2024-01-10
strom: added bbbvpn VLAN

Noki:


pmelange:
Is there a way to have it only generate an OLSRd config and no babel config?  Maybe give it only an ipv4 address.

pmelange:
*for VLAN 1312

pmelange:
when an interface doesn't have an ip6 address, should it maybe then look like this?
```
config interface
	option 'ifname' 'mesh_XXXXX'
	option 'ignore' 'true'
```

pmelange:
Interfaces mesh_perl and mesh_tu should probably be ```Mode 'mesh'```

pmelange:
Can we also get snmp statistics?

Noki:
> when an interface doesn't have an ip6 address, should it maybe then look like this?
> 
> ```
> config interface
> 	option 'ifname' 'mesh_XXXXX'
> 	option 'ignore' 'true'
> ```

I would keep it as is for now. If you want you can start a discussion about this in the bbb-configs channel. I don't really have an option about this.

PR #714, opened 2024-01-13
ak36: vlans and services

pktpls:


FFHener (approved):
Lgtm, but would still like to wait for a second review

pmelange:
The generated config looks good, except that interface mesh_bbbvpn is also in the babled config.  It doesn't need to be there at all as the bbbvpn doesn't speak babel. Commit https://github.com/freifunk-berlin/bbb-configs/commit/db1af477f31090016ef6af0bf0f6d4a9a7893c53 should prevent the babeld config section from being created.

locations/ak36.yml
```diff
@@ -77,70 +74,63 @@ mgmt:
...
     ipv6: 2001:bf7:750:4001::4/128
     metric: 1024
     ptp: true
 
-  - name: mesh_backbone
-    ifname: eth6
+  - name: mesh_bbbvpn
+    ifname: eth1.198
     ipv4: 10.31.130.164/32
     ipv6: 2001:bf7:750:4001::5/128
```
Noki, line 116: please make IPv6 a comment as we've done with https://github.com/freifunk-berlin/bbb-configs/blob/master/locations/strom.yml#L82 so no babeld config will be generated for this interface

PR #716, opened 2024-01-16
f17: converted to single file location format

FFHener:
@imaginator could you review this changes? I converted the config file to the new single file format. There shoudn't be any changes to the config.


Noki:
Except core, emma and zwingli, all other hosts are offline. This config should be adjusted to match the current state.

pktpls:
LGTM :+1: 

> Except core, emma and zwingli, all other hosts are offline. This config should be adjusted to match the current state.

The location maintainer @imaginator can take care of this is separately

imaginator:
@Noki sorry for the delay at looking at this. I'll pull and rebuild off the new single file setup. The file looks pretty sensible.

PR #717, opened 2024-01-16
grueni: converted to single file location format

FFHener:


Noki:
@robertfoss I think this location ist grueni73 and this location can be deleted, right?

pktpls:
LGTM :+1: @robertfoss can delete it if neccessary

FFHener:
I wrote a pm to @robertfoss to follow up this topic

PR #721, opened 2024-01-16
kub: converted to single file format

FFHener:


locations/kub.yml
```diff
@@ -1,4 +1,28 @@
...
+
+snmp_devices:
+  - hostname: kub-simeon
+    address: 10.31.139.2
+    snmp_profile: airos_8
+
+
 ipv6_prefix: "2001:bf7:830:8800::/56"
 
 # Rio-Reiser got following prefixes:
```
Noki, line 28: this looks like an artefact from copying a config. might be good to change this in the process.

pktpls:
LGTM :+1: I'll remove that stray comment separately

pktpls:
> I'll remove that stray comment separately

Uff, branch protection, I love it...

locations/kub.yml
FFHener: I made a Issue #725

PR #739, opened 2024-01-27
snmp_profiles: add SwOs and Edgeswitch

FFHener:
This PR brings monitoring for Mikrotik SwitchOS and Ubiquitys Edgeswitches. In addition it fixes the monitoring of proprietary devices for our gateways.

- [x] Temperatur for SwOs
- [x] Uptime
- [x] Port Status
- [x] Add external SNMP monitoring to gateway devices (not rolled out yet as i didnt wanted to have unnecessary downtime at gateways)
- Fix Uptime Scale for Grafana freifunk-berlin/meta/issues/29
- Better Display Port Status in Grafana freifunk-berlin/meta/issues/29
- Better Display Port Speed in Gradana freifunk-berlin/meta/issues/29

### DIFF
--- a/group_vars/all/snmp_profiles.yml
+++ b/group_vars/all/snmp_profiles.yml
@@ -179,3 +179,99 @@ collectd_snmp_profiles:
       TypeInstance: "TX Bitrate"
       Scale: 1000000
       Values: .1.3.6.1.4.1.14988.1.1.1.8.1.13.1
+
+  swos:
+    sys_uptime:
+      Type: uptime
+      TypeInstance: days
+      Table: false
+      Scale: 1.15740740741e-07
+      Values: .1.3.6.1.2.1.1.3.0
+
+    temp:
+      Type: temperature
+      TypeInstance: Celsius
+      Table: false
+      Scale: 0.1
+      Values: .1.3.6.1.4.1.14988.1.1.3.11.0
+
+    if_speed:
+      Type: bitrate
+      TypeInstanceOID: .1.3.6.1.2.1.17.1.4.1.1
+      Scale: 1000000
+      Table: true
+      Values: .1.3.6.1.2.1.2.2.1.5
+
+    if_oper_status:
+      Type: gauge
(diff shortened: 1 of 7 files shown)


FFHener:
As the grafanapart is independent from this part i moved this part into a new meta issue. We could discuss as well if we should add snmp-monitoring for all switches treewide.

FFHener:
Some SwOS-Switches dont provide temperature data resulting in this error in logread:
`daemon.err collectd[9583]: snmp plugin: OID "iso.3.6.1.4.1.14988.1.1.3.11.0" is undefined (type ASN_NULL)`
I personally can live with that right now. When we collected more experience with the monitoring of the different models of switches we could then move over to more individualized snmp-profiles, maybe in combination with #751 (defining switches as hosts and match the corresponding snmp-profile via the model).

FFHener:
I found another error. At `fesev-switch` collectd reported `daemon.err collectd[2546]: uc_update: Value too old: name = fesev-switch/snmp/gauge; value time = 1706777397.464; last cache update = 1706777397.464;`. The Switch has the Corerouter as the time server, so this needs to be further investigated.

FFHener:
The Problem is solved.

TLDR: Names of the value descriptors wherent unique.

I was using the port names as descriptors for the individual names because i thought this would be more helpful in understanding the graphs in grafana (e.g. `zwingli-sama` helps more than `7`). The problem is that port names sometimes get used multiple times e.g. `dhcp`, `intern`, ..., but collectd uses these names from my understanding as keys. This problem seems kind of commen as it is mentioned in there [docs](https://github.com/collectd/collectd/wiki/Plugin-SNMP). After switching the names in `TypeInstanceOID` to Port Numbers like `N` or `GigabitEthernetN` the problem went away. So this PR is ready to be merged now.

FFHener:
The Error regarding temperature values is fixed as i introduced a new `snmp_profile` for SwOS Lite.

Got Out of Band approval from @pktpls

Issue #749, opened 2024-01-29
Modelfiles: Add nicer naming options


FFHener:
For nicer looking automatically generated documentation I would like to add the following attributes to each model file:

- vendor
- model
- version (if present)

I would add "dummy" modelfiles for proprietary devices as well so we can already start using the nicer names and use them later one to add more infos for generating config files for them

PR #750, opened 2024-01-29
Treewide: Add nicer names to modelfiles

FFHener:
This is related to #749

### DIFF
--- a/group_vars/model_ubnt_edgerouter_x_sfp.yml
+++ b/group_vars/model_ubnt_edgerouter_x_sfp.yml
@@ -1,5 +1,7 @@
 ---
 target: ramips/mt7621
+brand_nice: Ubiquiti
+model_nice: EdgeRouter X SFP
 
 dsa_ports:
   - eth0
--- a/group_vars/model_ubnt_nanostation_ac_loco.yml
+++ b/group_vars/model_ubnt_nanostation_ac_loco.yml
@@ -1,6 +1,8 @@
 ---
 target: ath79/generic
 openwrt_version: 22.03-SNAPSHOT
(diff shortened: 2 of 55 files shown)


group_vars/model_ubnt_edgerouter_x_sfp.yml
```diff
@@ -1,5 +1,7 @@
 ---
 target: ramips/mt7621
+brand_nice: Ubiquiti
+model_nice: EdgeRouter x-SFP
```
Noki, line 4: should be EdgeRouter X SFP

group_vars/model_ubnt_nanostation_ac_loco.yml
```diff
@@ -1,6 +1,8 @@
 ---
 target: ath79/generic
 openwrt_version: 22.03-SNAPSHOT
+brand_nice: Ubiquiti
+model_nice: Nanostation AC loco
```
Noki, line 5: should be "Loco"

group_vars/model_x86_64.yml
```diff
@@ -1,8 +1,8 @@
 ---
-
 target: x86/64
 override_target: generic
 image_search_pattern: "*-ext4-combined.img*"
+model_nice: Virtual Machine
```
Noki, line 5: this can also be bare metal

Noki (changes requested):

group_vars/model_x86_64.yml
FFHener: What do you think about the new name?

Noki (approved):

PR #752, opened 2024-01-30
mahalle: converted to single file location

noxilixon:
I will add mac-overwrite and suggested owm-data in an extra pr.


FFHener:
Although this is out of scope for converting to a single file location you would make me happy if you would include the height attributes from the wikipage
> Hoehe = 2
ueberNN = 31

as
```yml
altitude: 31
height: 2
```
in the owm-data section of the config file.

FFHener (approved):
lgtm

PR #760, opened 2024-02-05
Treewide: Add snmp monitoring for some switches

FFHener:
This PR adds Monitoring for some switches and several new profiles to archive this. These changes got already deployed at the gateways. Grafana/cgp arent yet at the state of showing all of these values in a beautifull way but this shouldnt stop us from slowly rolling out more data collection from the switches.

Although i didnt touched it Im unhappy about the current `airos_6` Situation. When no Station is connected  to a AirOS 6 AP the snmp-call for the name of the connected stations obviously cant work resulting in a NoSuchName-Error in our logs. This means that we get at e.g. `chris-core` every 30 seconds 30 errors. We should address this in the future. #766 

TODOs:
- [x] Add Profile for tplink
- [x] Add Profile for edgerouter
- [x] Add + Deploy Monitoring for ak36
- [x] Add + Deploy Monitoring for l105
- [x] Add + Deploy Monitoring for ohlauer
- [x] Add + Deploy Monitoring for saarbruecker
  - [x] Remove saarbruecker-sama monitoring from sama
  - [x] ~~Add profile for Gigabeam LR and add monitoring (also on segen)~~ use airos_8 profile as the missing data doesnt get reported. We get zeros as a reply which is handable.
- [x] Add + Deploy Monitoring for strom
- [x] Investigate the NoSuchName Error on airos_6 devices -> Solution #766 
  - [x] Investigate if we can compress this error
  - [x] ~~If not~~ write a comment
- [x] Check Scales -> need to be fixed somewhere else. The wrong scale only happens at Grafana. This is because the script feeding prometheus doesn't hand over the scale Information's if i understood rtznmpfl correctly
- [x] Make shure that snmp is correctly configured on all ~~hosts~~ gateway devices

### DIFF
--- a/locations/l105.yml
+++ b/locations/l105.yml
@@ -1,5 +1,4 @@
 ---
-
 location: l105
 location_nice: Lützowstraße 105
 latitude: 52.502040
@@ -8,13 +7,27 @@ altitude: 62
 community: true
 
 hosts:
-
   - hostname: l105-gw
     role: gateway
     model: "x86-64"
     image_search_pattern: "*-ext4-combined.img*"
     l105__disabled_services__to_merge:
       - "bird"
+
+snmp_devices:
+  - hostname: l105-poe
+    address: 10.31.127.130
+    snmp_profile: edgeswitch
+
+  - hostname: l105-rhxb
+    address: 10.31.127.131
+    snmp_profile: af60
+
+  # tub is not existent as of 02/24
+  # - hostname: l105-tu
+  #   adress: 10.31.127.132
+  #   snmp_profile: af60
+
 
 # L105 got following prefixes:
 # Router: 10.31.127.128/25 2001:bf7:750:3f00::/56
@@ -38,9 +51,9 @@ mgmt:
   ipv6: 2001:bf7:750:3f00::/64
   assignments:
     l105-gw: 1    # .129
-    l105-poe: 2   # .130 poe
-    l105-rhxb: 3  # .131 mesh_rhxb
-    l105-tu: 4    # .132 mesh_tu
+    l105-poe: 2   # .130
+    l105-rhxb: 3  # .131
+    l105-tu: 4    # .132
 
 # Mesh Network: 10.31.127.160/27
 mesh_links:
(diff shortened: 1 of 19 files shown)


locations/l105.yml
```diff
@@ -38,9 +51,9 @@ mgmt:
   ipv6: 2001:bf7:750:3f00::/64
   assignments:
     l105-gw: 1    # .129
-    l105-poe: 2   # .130 poe
-    l105-rhxb: 3  # .131 mesh_rhxb
-    l105-tu: 4    # .132 mesh_tu
+    l105-poe: 2   # .130
+    l105-rhxb: 3  # .131
+    l105-tu: 4    # .132
```
Noki, line 56: this should be removed or changed into a comment

locations/saarbruecker.yml
```diff
@@ -1,17 +1,34 @@
...
 
+snmp_devices:
+  - hostname: saarbruecker-sw
+    address: 10.31.83.50
+    snmp_profile: edgerouter
+
+  # hds has no roof installation as of 02/24
+  # - hostname: saarbruecker-hds
+  #   address: 10.31.83.51
+  #   snmp_profile: af60
```
Noki, line 21: this device is up and running, even if it has no neighbor it should be included. But it's MikroTik, not AF60

locations/segen.yml
```diff
@@ -32,10 +32,14 @@ hosts:
     model: ubnt_nanostation-m2_xm
 
 snmp_devices:
-  # TODO: Create a Gigabeam LR SNMP profile and use it for segen-saarbr
-  # - hostname: segen-saarbr
-  #   address: 10.31.6.10
-  #   snmp_profile: airos_8
+  - hostname: segen-switch
+    address: 10.31.6.2
+    snmp_profile: edgerouter
```
Noki, line 37: this should be edgeswitch, not edgerouter

Noki (changes requested):

Noki (approved):
looks good to me

PR #763, opened 2024-02-05
newyorck: remove location

noxilixon:
This location was only temporary active over a year ago. There will be freifunk again in the near future. But this repository should represent the current status. Also the new installation will be different from the old one, keeping the old config files has no use.


FFHener (approved):
lgtm

PR #765, opened 2024-02-06
workflows: fix wikiupdate and bump version to v4

FFHener:
Move the wikiupdate-workflow into a independent file making it way more clean. And it is working now as intended (instead of building all locations) as you can see here
[with a changed location](https://github.com/FFHener/bbb-configs/actions/runs/7807563218/job/21296175674)
[without a changed location ](https://github.com/FFHener/bbb-configs/actions/runs/7807593320/job/21296262014)

I use `||true` so the task will never fail because there might be locations that doesn't get updated because they don't have a wikiarticle.

In addition i bumped the version of `actions/checkout@v3` to v4 as the old one was using Node.js 16 which is deprecated.

### DIFF
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -1,6 +1,6 @@
 ---
 name: Test Build
-on: [pull_request] # yamllint disable-line rule:truthy
+on: [pull_request]  # yamllint disable-line rule:truthy
 
 jobs:
   build:
--- a/.github/workflows/testbuild.yml
+++ b/.github/workflows/testbuild.yml
@@ -1,12 +1,6 @@
 ---
 name: Test Build
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+on: [pull_request] # yamllint disable-line rule:truthy
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/testbuild.yml
```diff
@@ -1,20 +1,14 @@
 ---
 name: Test Build
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+on: [pull_request]  # yamllint disable-line rule:truthy
```
Noki, line 3: I disagree with this. We want to do test builds after the merge because you could also merge stuff without doing a rebase first. Config builds are fast now anyway, so let's always do them.

.github/workflows/testbuild.yml
```diff
@@ -45,22 +39,14 @@ jobs:
...
-        run: |
-          ansible-playbook play.yml --tags wiki --limit "${CHANGED_HOSTS:+$CHANGED_HOSTS,}$CHANGED_LOCATIONS"
-        continue-on-error: true
-
-      - if: ${{ github.event_name == 'pull_request' }}
-        name: Store build output
+      - name: Store build output
         uses: actions/upload-artifact@v1
         with:
           name: ansibleimages
```
Noki, line 52: we don't build images anymore. we should change this to store the configs as artifacts instead

Noki (changes requested):
We should always build configs, not just on pull requests. We should store the configs as artifacts because we don't build images anymore.

.github/workflows/testbuild.yml
FFHener: Sounds good to me, i just reverted my initial commit where i added wikiupdater here, but i will add it back.

.github/workflows/testbuild.yml
FFHener: I dont have to much time right now to dig in to this, feel free to propose something. Otherwise i will take care when i have time again

.github/workflows/testbuild.yml
```diff
@@ -45,23 +45,15 @@ jobs:
...
-
-      - if: ${{ github.event_name == 'pull_request' }}
-        name: Store build output
+      - name: Store build output
         uses: actions/upload-artifact@v1
         with:
-          name: ansibleimages
-          path: ./tmp/images
+          name: ansibleconfigs
+          path: /tmp/config
```
Noki, line 59: that's not working...

FFHener: I know, was just a quick test.
>I dont have to much time right now to dig in to this, feel free to propose something. Otherwise i will take care when i have time again

FFHener: It annoyed me to have this PR open so i fixed it. In addition i renamed everything to account for the change of the task

Noki (approved):
looks good to me

PR #770, opened 2024-02-09
spitta13: prefer zwingli as uplink

FFHener:


Noki:
I don't really understand this PR. The nearest gateway is always via zwingli, so where else would it go? Usually we only set lqm on specific mesh interfaces if there is a real issue and I don't see it in this case. It might however be useful to set an lqm on 2.4 GHz so kaskel29-nord prefers 5 GHz. In addition to lqm you should also set mesh_metric for babeld at the same time.

FFHener:
The route without lqm goes via kasekl29. This was likely not visible as I already added these changes manually.
You are right about adding mesh_metric, i will add this.

```
root@spitta13-core:~# mtr 1.1.1.1 -r
Start: 2024-02-10T17:15:08+0100
HOST: spitta13-core               Loss%   Snt   Last   Avg  Best  Wrst StDev
  1.|-- 10.31.78.138               0.0%    10    5.2  16.0   2.5  93.1  28.9
  2.|-- 172.31.224.1               0.0%    10   21.5  22.1  18.8  31.3   3.5
  3.|-- host01-br.wandale0.hambur  0.0%    10   21.2  25.6  19.2  46.4   8.3
  4.|-- 100.64.112.17              0.0%    10   21.3  27.3  19.4  52.7  11.4
  5.|-- ge-1-1-7.cr01.ham01.lwlco  0.0%    10   25.7  33.1  17.5  67.9  17.0
  6.|-- 162.158.200.2              0.0%    10   20.5  28.4  20.5  40.8   7.5
  7.|-- one.one.one.one            0.0%    10   22.4  23.1  19.6  30.6   3.6

```

PR #772, opened 2024-02-10
fffw-marktplatz: init location

Noki:


Noki:
@akira25 will this be used or should we just close it?

PR #774, opened 2024-02-10
Add snmp-profile routeros and add monitoring to fanninger47

FFHener:


group_vars/all/snmp_profiles.yml
```diff
@@ -277,7 +311,7 @@ collectd_snmp_profiles:
       Table: true
       Values: .1.3.6.1.2.1.2.2.1.14
 
-  edgerouter:
+  edgeBei:
```
Noki, line 314: accident?

FFHener: Thanks for spotting

PR #775, opened 2024-02-10
workflows: proper loading of wikiupdater secrets

FFHener:


Noki (approved):
looks good to me, except the agym change, that does not need to be in this commit

PR #778, opened 2024-02-14
mela: 2g uplink, second core until install is fixed

Noki:
This PR recovers the second part of the mela installation. The old core router is still broken and need to be replaced. Due to the fact that we have 2 switches the installation is now split in half.

### DIFF
--- a/locations/mela-2g.yml
+++ b/locations/mela-2g.yml
@@ -1,0 +1,180 @@
+---
+location: mela-2g
+location_nice: Melanchthonkirche, Melanchthonplatz, 13595 Berlin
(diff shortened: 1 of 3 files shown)


FFHener:
Is torte-mela-2g a temporary installation?

Noki:
> Is torte-mela-2g a temporary installation?

At least I consider it temporary. the link is really bad and once the mela-core is replaced we don't need it anymore and should turn it off again. I just pushed it so we have the current state of everything related to mela.

FFHener (approved):
Lgtm

PR #780, opened 2024-02-17
zwingli: add packages to manage meshtastic node

Noki:
We added a meshtastic node to zwingli and this adds the packages required to mange it remotely.
The artifact upload is broken due to files with colons, the workflow needs to be adjusted / fixed.


FFHener (approved):
Lgtm

PR #783, opened 2024-02-20
xdorf: replace core router

noxilixon:
This fixes freifunk-berlin/meta#44.

I switched the device, because I had a 4040 available. It is flashed and pwd is set, ready to be replaced as soon as I get contact to people from the location.


FFHener (approved):
lgtm, will convert it back to a draft. We can merge it when the 4040 is deployed

PR #788, opened 2024-02-21
k12-haus3: add webcam as olsrd service

Noki:


FFHener:
Isn't #524 supposed to resolve this in a cleaner and more general approach?

Noki:
yes, but we agreed that we want to add everything as hosts, so that PR needs to be completely reworked after that and this gets the job done for now. ;-)

FFHener (approved):
lgtm

PR #790, opened 2024-02-25
Docs: Added missing options, fixed typos, reduced duplications

FFHener:


DEVELOPER.md
```diff
@@ -98,23 +90,17 @@ airos_dfs_reset:
...
-  - name: "segen-mabb"
-    target: "10.31.6.12"
-    username: "ubnt"
-    password: "file:/root/pwd"
-    daytime_limit: "2-7"
 ```
 
 ### network
 
 This part defines IP-Addresses, WIFI-Properties, Mesh and so on. Lets have a (shortened) example of a `networks.yml` from the magda-location:
```
Noki, line 97: we should remove the networks.yml reference here as we now have single file format

Noki (approved):
looks good to me, besides the remaining network.yml reference

PR #794, opened 2024-02-27
kirchhof: change prdhcp ssid

noxilixon:
Is that enough? Do we need to change more of the wireless profile?


pktpls:
Yep that's enough :+1:

Issue #715, opened 2024-01-14
DAWN: communication between devices does not work properly


Noki:
After the startup of the core router / ap the luci web interface only shows the core router / ap and not the other wireless devices. Due to missing communication between devices band steering / roaming does not work.

![image](https://github.com/freifunk-berlin/bbb-configs/assets/198902/9bd61d2f-2c38-421f-9fc2-5c334ea5e7f5)

To to solve this problem until the next reboot it is sufficient to restart umdns (`/etc/init.d/umdns restart`). Once restarted on one device, the device will show up in the list of the other devices:

![image](https://github.com/freifunk-berlin/bbb-configs/assets/198902/42f568ee-4e43-4e43-97cd-0aaef32d986b)

Noki:
This seems to be fixed by now, at least it's working again for me for some time now.

PR #815, opened 2024-04-15
radbahn: init location

pktpls:


locations/radbahn.yml
```diff
@@ -0,0 +1,93 @@
...
+hosts:
+
+  - hostname: radbahn-core
+    role: corerouter
+    model: ubnt_usw-flex
+
+  - hostname: radbahn-n-nf
+    role: ap
+    model: zyxel_nwa55axe
+    openwrt_version: snapshot
```
Noki, line 20: Why do we need snapshot here? 23.05-snapshot is working just fine...

Noki (commented):
In my opinion the APs should both use the standard channels and and support meshing on at least2.4 GHz, better also on 5 GHz so Falter nodes can connect without any changes. In addition wireless roaming should be activated.

Noki:
> Roaming ist still buggy, right? #715 I think having different channels make sense to avoid interference for the Clients?

Roaming is fixed and works for me now. I closed the issue. 

In general having differen channels is great, however if you do not offer a default channel for meshing you will not see new Falter nodes connect. In addition I don't think interference is much of a problem for this location because the distance between the APs is quite large compaired to other locations.

locations/radbahn.yml
pktpls: To my knowledge mt7915e meshing f'cked in 23.05, no?

Noki: works fine for me

locations/radbahn.yml
```diff
@@ -0,0 +1,103 @@
...
+  - vid: 42
+    name: mgmt
+    role: mgmt
+    prefix: 10.31.251.0/26
+    ipv6_subprefix: 1
+    gateway: 1
+    dns: 1
+    assignments:
+      radbahn-core: 1
+      radbahn-emmaus: 2
```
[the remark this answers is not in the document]
pktpls: Ich finde es schräg wenn Männer ihren Technikspielzeugen Frauennamen geben, aber das ist ein anderes Thema, habe es geändert

PR #825, opened 2024-04-26
ekke: init

FFHener:
I know about the current problems of the core router model, but the location is already deployed.

### DIFF
--- a/locations/ekke.yml
+++ b/locations/ekke.yml
@@ -1,0 +1,128 @@
+---
+
+location: ekke
+location_nice: Ekkehardstr. 6
+latitude: 52.46809
+longitude: 13.48709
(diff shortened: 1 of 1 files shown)


locations/ekke.yml
```diff
@@ -0,0 +1,130 @@
...
+
+      - mode: ap
+        ssid: Interwebz
+        encryption: psk2
+        key: 'file:/root/wifi_pass'
+        network: uplink
+        radio: [11a_standard, 11g_standard]
+        ifname_hint: uplink
+
+ssh_keys: # to avoid adding the default keys
```
pktpls, line 128: just do `ssh_keys: []` here

pktpls (approved):
LGTM, one small comment

PR #832, opened 2024-05-01
utilities: add mass update script

Noki:
This scripts aims to improve mass updates. The script checks for availability, free memory in comparison with the image size and a successful file transfer.

Note: This script will skip some low memory devices and could be improved so it could also flash these in the future, but for now I did not really care.

### DIFF
--- a/mass-update.sh
+++ b/mass-update.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+
+# Define file directory and endings
+FILE_DIR="tmp/images"
+FILE_ENDINGS=".itb .bin"
+
+# Find files matching the specified endings
+FILES=""
+for ENDING in $FILE_ENDINGS; do
+    FILES="$FILES $(find "$FILE_DIR" -type f -name "*$ENDING")"
+done
+
+# Sort files based on whether filename contains "core" or not
+CORE_FILES=""
+OTHER_FILES=""
+for FILE_PATH in $FILES; do
+    if [[ "$FILE_PATH" == *"core"* ]]; then
+        CORE_FILES="$CORE_FILES $FILE_PATH"
+    else
+        OTHER_FILES="$OTHER_FILES $FILE_PATH"
+    fi
+done
+SORTED_FILES="$OTHER_FILES $CORE_FILES"
+
+# Print information and prompt for confirmation
+echo ""
+echo "This script will do the following:"
+echo ""
+echo "- flash all the following hosts with the corresponding firmware files currently present in $FILE_DIR"
+echo "- first flash APs, than core routers based on the naming convention"
+echo "- check the availability of the hosts before and after flashing"
+echo "- ignore keychecking"
+echo "- make sure that at least 16 MB of RAM are available before performing a sysupgrade"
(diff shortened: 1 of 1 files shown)


mass-update.sh
```diff
@@ -0,0 +1,122 @@
...
+        ssh-keyscan "$HOSTNAME" >> ~/.ssh/known_hosts
+
+        # Check memory on remote host
+        MEMORY=$(ssh "root@$HOSTNAME" "free | awk 'NR==2 {print \$4}'")
+        if [ "$MEMORY" -ge $(( $(stat -c %s "$FILE_PATH") / 1024 + 3072 )) ]; then  # File size in KB + 3 MB
+            echo "Memory on $HOSTNAME is sufficient ($MEMORY KB)"
+
+            # SCP the file
+            echo "Copying $FILENAME to $HOSTNAME:/tmp/"
+            if scp "$FILE_PATH" "root@$HOSTNAME:/tmp/"; then
```
pktpls, line 85: `scp -O` if the destination host doesn't have `openssh-sftp-server` installed

mass-update.sh
```diff
@@ -0,0 +1,122 @@
...
+                # Wait for hostname to become unreachable
+                echo "Waiting for $HOSTNAME to become unreachable..."
+                while ping -c 1 "$HOSTNAME" >/dev/null 2>&1; do sleep 1; done
+
+                # Wait for 20 seconds before checking hostname reachability again
+                sleep 20
+
+                # Debug output: Waiting for hostname to become reachable again
+                echo "Waiting for $HOSTNAME to become reachable again..."
+                while ! ping -c 1 "$HOSTNAME" >/dev/null 2>&1; do sleep 1; done
```
pktpls, line 100: `ping -c` can be combined with `ping -w` so you don't need the loop

mass-update.sh
```diff
@@ -0,0 +1,122 @@
...
+    echo "Hostname: $HOSTNAME"
+
+    # Check if hostname is reachable
+    echo "Checking if $HOSTNAME is reachable..."
+    if ping -c 1 "$HOSTNAME" >/dev/null 2>&1; then
+        echo "Hostname $HOSTNAME is reachable"
+
+        # Removing old host key
+        echo "Removing old host key for $HOSTNAME"
+        ssh-keygen -R "$HOSTNAME" >/dev/null 2>&1
```
pktpls, line 72: You can circumvent all the host key stuff: `sshnocheck='ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'`

pktpls:
LGTM, small comments, `scp -O` is the only important one

mass-update.sh
Noki: The idea of the loop is to be be able to wait for quite some time and to be stuck if the device does not come back. Exiting after a deadline e.g. 120s would of course be better at that point. I don't get how to achieve this with the combination of `-c` and `-w` as `ping -c 1 -w 120 nonexisting.olsr` exits immediately with a "Name or service not known" error.

PR #840, opened 2024-05-20
chris: use default channel for 802.11s antennas

Noki:
We should use the default channel for 2.4 GHz 802.11s links so new falter nodes with default config could mesh.

Note: I just fixed the VLANs at chis-core as all the 802.11s VLANs were missing on the switch.

### DIFF
--- a/locations/chris.yml
+++ b/locations/chris.yml
@@ -171,8 +171,9 @@ networks:
       chris-s-2ghz: 12
       chris-w-2ghz: 13
 
+# Use default channel for all 802.11s links so standard falter nodes can mesh
 location__channel_assignments_11g_standard__to_merge:
-  chris-n-2ghz: 1-20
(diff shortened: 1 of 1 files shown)


Noki:
There is almost no dhcp traffic, so I don't think there will be much of a difference: https://monitor.berlin.freifunk.net/cgp/detail.php?p=interface&pi=br-dhcp&t=if_octets&h=chris-core&s=31536000&x=800&y=350

PolynomialDivision:
> I see your point, but this could make the local experience for clients worse, right?

Yes. This leads to more interference. Also channel 13 sometimes does not work with some legacy devices. Also maybe a controversial point is that for me it is more likely that we have people in 2.4 ghz dhcp than meshing falter nodes.

Noki:
ok, so we leave it as is and optimize for dhcp user experience. fine for me.

PR #844, opened 2024-05-25
weidenbaum: init location

Noki:
weidenbaum-r0 is connected via LAN but not setup as a standalone router that meshes via LAN. The reason for this is that we had issues with flashing Ubiquiti UniFi AC Mesh devices and bricked them in the process. We did not want to risk this today. This might be changed in the future. We also keep the mesh on the core router active as a backup, but make the metrics worse so routing via weidenbaum-r0 is preferred. ~~There might also be an opportunity to connect to chris-core, but we had no antenna at hand to test this out, but already prepared the configuration to test this out.~~


Noki:
comment added

PR #850, opened 2024-06-04
magda: liese-11-sw-core lqm

Noki:


Noki:
> Is there a way we could use hostnamens instead of IPs?

The format is `option LinkQualityMult ‘<remote mesh interface ip> <multiplier>’` so to my understanding this does not work with hostnames.

PR #854, opened 2024-06-10
f17: goodbye

Noki:
A couple of weeks ago Simon told me that he is leaving Berlin and f17 will therefore shut down. The node is offline now for a couple of days. @imaginator please confirm that this is correct and please make sure to give back the IPs to the pool (link was in the registration mail / can be requested again via https://config.berlin.freifunk.net/summary)


imaginator:
@Noki thanks for checking this. Yes, I can confirm the antennas went down today (https://twitter.com/imaginator/status/1800138148804219378) and I've unregistered the IPs. 

So long and thanks for all the packets.

PR #856, opened 2024-06-12
rhnk: replace core with thinkcenter; remove outdated firmware versions

Noki:


Noki:
https://github.com/freifunk-berlin/bbb-configs/pull/883 replaces this one

PR #860, opened 2024-06-18
segen: fix names

Noki:
Reflect the actual orientation of the antenna


Noki:
> We should update the name in airos as well

that was already done long time ago...

PR #862, opened 2024-06-18
all: update the upstream DNS servers

pktpls:
- Add Quad9 "recommended" servers (malware blocking enabled, DNSSEC validation enabled).
- Remove AS250.net which has been broken at L105 for a long time.
- Remove digitalcourage.de, they offer only DNS-over-TLS, which we're not prepared for at the moment.

dnsmasq picks random upstream server for every incoming DNS request. The timeout before it tries the next seems to be around 5 seconds.

Removing the broken upstream servers will improve UX significantly.

### DIFF
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -2,15 +2,13 @@
 zonename: 'Europe/Berlin'
 timezone: 'CET-1CEST,M3.5.0,M10.5.0/3'
 
+# TODO: find a second good DNS upstream in Berlin
 dns_servers:
-  # dns3.digitalcourage.de @ hetzner falkenstein
-  - 2a01:4f8:251:554::2
-  - 5.9.164.112
-  # ns1.fdn.fr @ gitoyen paris
-  - 2001:910:800::40
-  - 80.67.169.40
-  # dns.as250.net anycast (l105 broken)
-  # - 194.150.168.168
+  # quad9.net @ megaport l105+ak36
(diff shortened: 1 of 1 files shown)


spolack:
I'd also watch out for latency as this significantly contributes to snappiness and try to prefer anycasted resolvers with a served in Berlin.

pktpls:
> I'd recommend to watch out for latency as this significantly contributes to snappiness and try to prefer anycasted resolvers with a served in Berlin.

Is that a +1 or a request for changes? :-)

pktpls:
I suppose we could also debug as250.net for a second anycast DNS upstream. It generally works, but via l105 gateway it loops (`mtr 194.150.168.168`).

pktpls:
I have contacted Artikel 10 e.V. to ask if we can use their public resolver. It can do IPv4 and IPv6 and is hosted at IN-Berlin / ak36.

pktpls:
So, we don't really have any good options in Berlin apart from Quad9, until we implement DNS-over-TLS (which isn't too complicated). I suggest we use only Quad9 temporarily, and artikel10.de will become an option once we have DNS-over-TLS.

The DNS UX is really bad right now: fdn.fr regularly shows signs of being overloaded, and digitalcourage isn't actually usable because it only supports DNS-over-TLS (sorry, my bad).

Have updated the PR to use only Quad9.

Noki (approved):

PR #864, opened 2024-06-19
Kotti hinzugefügt

kiangosling:


locations/kotti.yml
```diff
@@ -0,0 +1,69 @@
...
+    address: 10.248.2.2
+    snmp_profile: airos_8
+
+ipv6_prefix: 2001:bf7:830:6600::/56
+
+# Kotti got following prefixes:
+# Router: 10.248.2.0/26
+# --MGMT: 10.248.2.0/28
+# --MESH: 10.248.2.16/28
+# --DHCP: 10.248.2.32/27
```
Noki, line 29: Probably not enough IPs for a location like this.

locations/kotti.yml
```diff
@@ -0,0 +1,69 @@
+---
+
+location: kotti
+location_nice: Skalitzer Straße 134, 10999 Berlin
+latitude: 52.4992763
+longitude: 13.4195371
```
Noki, line 6: The location used to be differen. It this really mounted in a different spot now? If not, please correct:

image

locations/kotti.yml
```diff
@@ -0,0 +1,69 @@
...
+
+ipv6_prefix: 2001:bf7:830:6600::/56
+
+# Kotti has following prefixes:
+# Router: 10.31.167.208/28
+# --MGMT: 10.31.167.208/29
+# --MESH: 10.31.167.216/29
+# --DHCP: 10.248.2.0/26
+
+networks:
```
Noki, line 31: why don't we activate 802.11s mesh networks? I would activate them, so new neighbors could start meshing and get in contact.

PR #868, opened 2024-06-24
t1: init location

FFHener:
Everything is installed, just waiting for a permanent power solution


noxilixon (approved):
lgtm, but would remove the comment in last line

PR #869, opened 2024-06-24
forcki: removed

### LINKED ISSUE #782: forcki: readd location
It was in a private flat and seems like it wont come back to life

FFHener:
The IPs might need to be unregisterd @freifunkoperator. This fixes #782


pktpls:
I'll take notes on IP addresses to be freed up.

(Btw forcki-core was last seen online 2024-02-08)

PR #870, opened 2024-06-26
gub37: Add gub37-emma

robertfoss:


Noki:
creative ipv6_subprefix numbering... maybe just go with -[VID] ;-)

PR #890, opened 2024-07-16
sav: added switchconfig, changed corerouter

FFHener:
~~**EDIT: The second commit with the reload still needs to be fully tested**~~ -> removed 2nd commit to bring the current state into `main`, i will continue to investigate this with low priority as we can easily recover a device afterwards and this is highly custom solution that will eventually a feature in bbb-configs anyways

The requirement from the location is as you can see below, so i added a custom switch config based on @Akira25 's script for `ilr`.

Port 1: Connection to Emma
Port 2: DHCP for Clients
Port 3: Local Backup

image (1)

### DIFF
--- a/locations/sav.yml
+++ b/locations/sav.yml
@@ -17,19 +17,66 @@ contacts:
 ipv6_prefix: 2001:bf7:830:ae00::/56
 
 hosts:
-
   - hostname: sav-core
     role: corerouter
-    model: x86-64
+    model: "mikrotik_routerboard-750gr3"
+    host__rclocal__to_merge:
+      - '#'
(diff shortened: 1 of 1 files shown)


locations/sav.yml
```diff
@@ -17,19 +17,69 @@ contacts:
 ipv6_prefix: 2001:bf7:830:ae00::/56
 
 hosts:
-
   - hostname: sav-core
     role: corerouter
-    model: x86-64
+    model: "x86-64"
+    # model: "mikrotik_routerboard-750gr3"
```
pktpls, line 23: This needs to be switched

pktpls:
One comment, looks good otherwise :+1:

PR #894, opened 2024-07-16
hacrafu-barbine: init

ZuquetuPlexis:


pktpls:
You can define your custom wifi profile with `location__wireless_profiles__to_merge` - example: https://github.com/freifunk-berlin/bbb-configs/blob/main/locations/radbahn.yml#L117

If you use the same profile at multiple locations, you can add it to `group_vars/all/wireless_profiles.yml

PR #896, opened 2024-07-18
Mac override fixes

Noki:
see https://github.com/freifunk-berlin/meta/issues/94
one of the zwingli antennas has an unknown password. I will add an todo to the maintenance issue.


Noki:
These changes are already flashed to the corresponding devices.

Issue #901, opened 2024-07-18
linter: check for IPv6 duplicates


Noki:
- [x] identify all duplicates
- [x] fix all duplicates
  - [x] weidenbaum
  - [x] perle
  - [x] spitta13 / fanninger47
- [x] develop some kind of linter action to prevent this from happening again (#905)

PR #905, opened 2024-07-22
workflows: add ip prefix duplicate check / fix duplicates

Noki:
This ensures that we do not have duplicate IP prefixes any more:

image

closes https://github.com/freifunk-berlin/bbb-configs/issues/901

### DIFF
--- a/.github/checks/check-ip-prefix-duplicates.sh
+++ b/.github/checks/check-ip-prefix-duplicates.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Change to the locations directory
+cd locations || exit 1
+
+# Check for IPv4 duplicates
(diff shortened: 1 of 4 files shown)


Noki:
> Thank you for writing this check. It feels kind of hacky to use a bash-script with a regex. Could there be a cleaner way using ansible tasks instead as it interprets everything as variables and not just as a random text file?

Using ansible would be cleaner, but that's nothing I can do right now. Feel free to add an issue to rewrite this later on when someone has time. For now any check is better than no check.

This is ready for review and to be merged.

pktpls:
This is the kind of linter I like :) can be run locally and solves a real problem :+1: 

Also add them to `make lint` when all are merged.

PR #906, opened 2024-07-22
Ska95 fixes

FFHener:


locations/ska95.yml
```diff
@@ -19,31 +19,38 @@ dns_servers:
...
 # --DHCP: 10.36.8.64/26
+# --FREE: 10.31.8.32/27
 
 ipv6_prefix: "2001:bf7:830:700::/56"
 
 hosts:
   - hostname: ska95-core
     role: corerouter
     model: tplink_tl-wdr3600-v1
-    wireless_profile: freifunk_default
```
Noki, line 33: without this the core router will noch provide wifi any more. is this intended?

PR #907, opened 2024-07-22
workflows: add hostname duplicate check / fix duplicates

Noki:
This PR introduces a check to detect duplicate hostnames and fixes existing issues:

image


pktpls:
Love that the linter finds issues right away :+1:

PR #915, opened 2024-07-24
weidenbaum: make weidenbaum-r0 an AP

Noki:
not flashed yet. this requires a two step flash process via a minimalistic image.


pktpls:
Looks good :+1: 

Btw "ap" images are small, you don't need the special procedure if it's smaller than 7744k. For example dragonkiez-rathausblock-miami-ap1 is ~6850k.

pktpls (approved):

PR #919, opened 2024-07-25
Collectd: Adding ping monitoring for IPv6 

FFHener:
Right now we only monitor IPv4 pings, but I think IPv6 would interesting as well especially because we use different routing protocols for them. The IPv6 address is from Cloudflare as well to make them comparable.


Noki:
This is a nice addition!

PR #931, opened 2024-08-04
fardf: init location

Noki:
This PR is a first draft for converting fardf to bbb-configs based on the current setup, not following the bbb-configs conventions. The main motivation for this PR is to solve the instability issues with the current Falter firmware running at the core.

TODO:

- [x] Get a new IPv6 prefix and configure it
- [x] Get a new IPv4 prefix that could be used, split it and use it for MGMT during migration
- [x] Check if there are any special config settings that need to be preserved (found only outdated stuff)
- [x] Prepare SNMP and Airos DFS reset configuration
- [x] Create additional VLANs at the switch
- [x] Have at least one person review this
  - [x] Check prepared config and compare it to the existing setup; Are there any issues?

The migration would be as follows:

- [x] Switch all antennas to the new management VLAN and ip-addresses including dns, gateway and NTP
- [x] Change the switch to use the new management VLAN and ip-addresses including dns, gateway and NTP
- [x] Build and flash a minimal image for the core and flash it <= ~7 MB so partition resize could take place
- [x] Place pwd file into root directory
- [x] Reconfigure antennas to use the new Mesh VLANs
- [x] Remove VLAN 101 from configuration
- [x] Build a normal image for the core and flash it so we get full functionality
- [x] Remove old VLANS from switch

After bbb-configs migration:

- [x] Update UISP
- [x] Unregister old IP prefixes
- [x] Update Wiki

### DIFF
--- a/locations/fardf.yml
+++ b/locations/fardf.yml
@@ -1,0 +1,198 @@
+---
+location: fardf
+location_nice: "Finanzamt Reinickendorf, Eichborndamm 208, 13403 Berlin"
+latitude: 52.5870976
+longitude: 13.324892521
+altitude: 75
+community: true
+
+hosts:
+  - hostname: fardf-core
+    role: corerouter
+    model: "ubnt_unifiac-mesh"
+    wireless_profile: freifunk_default
+
+snmp_devices:
+  - hostname: fardf-switch
+    address: 10.248.11.130
+    snmp_profile: edgeswitch
+
+  - hostname: fardf-bht
+    address: 10.248.11.131
+    snmp_profile: airos_8
+
+  - hostname: fardf-maerkisches
+    address: 10.248.11.132
+    snmp_profile: airos_8
+
+  - hostname: fardf-sange
+    address: 10.248.11.133
+    snmp_profile: airos_8
+
+  - hostname: fardf-nord
+    address: 10.248.11.134
+    snmp_profile: airos_8
+
+  - hostname: fardf-ost
+    address: 10.248.11.135
+    snmp_profile: airos_8
+
+  - hostname: fardf-sued
+    address: 10.248.11.136
+    snmp_profile: airos_8
+
+  - hostname: fardf-west
+    address: 10.248.11.137
+    snmp_profile: airos_8
+
+airos_dfs_reset:
+  - name: "fardf-maerkisches"
+    target: "10.248.11.132"
+    username: "ubnt"
+    password: "file:/root/pwd"
+    daytime_limit: "2-7"
+
+  - name: "fardf-sange"
+    target: "10.248.11.133"
+    username: "ubnt"
+    password: "file:/root/pwd"
+    daytime_limit: "2-7"
+
+  - name: "fardf-nord"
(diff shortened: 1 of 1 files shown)


pmelange:
For the migration, it is necessary to do all the VLAN changes manually before the first mini-image flash.  Otherwise you will be locked out.

pmelange:
mesh interface "rudi" is not needed.  The device doesn't exist.  history: there was an attempt to make the connection to a garden, but it didn't work.

pmelange:
the contact info can be set to "community" and not specifically just me.

pmelange:
mesh interface BHT can be set to mesh mode "ether" as it's a client of the dish at BHT.

Noki:
Thanks for the feedback.

> mesh interface "rudi" is not needed. The device doesn't exist. history: there was an attempt to make the connection to a garden, but it didn't work.

removed it

> the contact info can be set to "community" and not specifically just me.

changed it

> mesh interface BHT can be set to mesh mode "ether" as it's a client of the dish at BHT.

set it to p2p

> For the migration, it is necessary to do all the VLAN changes manually before the first mini-image flash. Otherwise you will be locked out.

Hmm, I thought having the VLANs like I do would be a drop in replacement... So just to clarify:

Instead I should do:

- add missing VLANs to the switch (done)
- reconfigure each antenna to use the new VLAN 42 with the new IP/DNS/Gateway/NTP server for management
- reconfigure the switch to use the new VLAN 42 with the new IP/DNS/Gateway/NTP server for management
- change the bbb-configuration to directly reflect those changes
- flash the image

right?

btw,; the north antenna seems to be down and therefore can't be reconfigured.

pmelange:
> Hmm, I thought having the VLANs like I do would be a drop in replacement... So just to clarify:
> 
> Instead I should do:
> 
>     * add missing VLANs to the switch (done)
> 
>     * reconfigure each antenna to use the new VLAN 42 with the new IP/DNS/Gateway/NTP server for management
> 
>     * reconfigure the switch to use the new VLAN 42 with the new IP/DNS/Gateway/NTP server for management
> 
>     * change the bbb-configuration to directly reflect those changes
> 
>     * flash the image
> 
> 
> right?

I would leave the migration to you. I'm sure it'll work however you do it.  In the past, I and @PolynomialDivision migrated everything manually with the running image until it was ready for the bbbconfig image.  But like the saying goes, there's more than one way to skin a cat.

pmelange:
> btw,; the north antenna seems to be down and therefore can't be reconfigured.

That antenna is attached via sange (poe passthrough).  I can see it's on and that the link is at 1gbps.  But I also can't reach it.  Maybe someone has to go there and reset it and reconfigure it.

Noki:
This PR is ready for review. Maybe someone could find some time to doublecheck if all the VLANs are setup correctly, especially the pass through VLANs.

@pmelange please unregister the old IPs
@FFHener please make sure the wiki-updater updates the wiki based on this config

pmelange:
> @pmelange please unregister the old IPs

done

PR #934, opened 2024-08-09
Init location "FBLipke" and device "nsm2_xw"

LipkeGu:
This Pull Requests adds the Location and device "ubnt nsm2_xw"

### DIFF
--- a/group_vars/model_ubnt_nanostation_m2_xw.yml
+++ b/group_vars/model_ubnt_nanostation_m2_xw.yml
@@ -1,0 +1,29 @@
+---
+override_target: "ubnt_nanostation-m"
+target: ath79/tiny
+brand_nice: Ubiquiti
+model_nice: Ubiquiti Nanostation M2 (XW)
+version_nice: XW
+
+switch_ports: 3
+switch_int_port: 0
+switch_ignore_ports: [1, 2, 3, 4]
+int_port: eth0
+low_mem: false  #XW has 64MB
+
+dsa_ports:
+  - lan
+  - wan
+
+wireless_devices:
+  - name: 11g_standard
+    band: 2g
+    htmode_prefix: HT
+    path: platform/ahb/18100000.wmac
+    ifname_hint: wlan2
+    antenna_gain: 8
+
+poe_ports:
+  - name: PoE Passthrough
+    gpio_pin: 2
+    value: 0
--- a/locations/fblipke.yml
+++ b/locations/fblipke.yml
@@ -1,0 +1,75 @@
+---
+location: fblipke
+location_nice: FBLipke - Freifunk Berlin
+latitude: 52.484948320
+longitude: 13.443380903
+altitude: 42
+height: 13
+contact_name: 'Guido L.'
+contact_nickname: 'LipkeGu'
+contacts:
+  - 'contact@fblipke.de'
+hosts:
+  - hostname: fblipke-core
+    role: corerouter
+    model: "x86_64"
+    wireless_profile: freifunk_default
+    low_flash: true
+  - hostname: fblipke-ap01
+    role: ap
+    model: "ubnt_nanostation_m2_xw"
+    wireless_profile: freifunk_default
+    low_flash: true
+
+ipv6_prefix: "2001:bf7:850:1500::/56"
+
+networks:
+  - vid: 20
+    role: mesh
+    name: 11s_n_2g
+    prefix: 10.248.2.200/32
+    ipv6_subprefix: -20
+    mesh_ap: fblipke-core
+    mesh_radio: 11g_standard
+    mesh_iface: mesh
+  - vid: 40
+    role: dhcp
+    prefix: 10.248.3.128/27
+    ipv6_subprefix: 0
+    inbound_filtering: true
+    enforce_client_isolation: true
+    no_corerouter_dns_record: true
+    assignments:
+      fblipke-core: 129
+      fblipke-ap01: 130
+  - vid: 42
+    role: mgmt
+    prefix: 10.248.3.160/28
+    gateway: 1
+    ntp: 1
+    ipv6_subprefix: 1
+    assignments:
+      fblipke-core: 1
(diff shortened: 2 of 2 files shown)


locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+    wireless_profile: freifunk_default
+    low_flash: true
+  - hostname: fblipke-ap01
+    role: ap
+    model: "ubnt_nanostation_m2_xw"
+    wireless_profile: freifunk_default
+    low_flash: true
+
+ipv6_prefix: "2001:bf7:850:1500::/56"
+
```
Noki, line 25: please state all the ips of your location here so everybody can very. usually we register one large ipv4 range and one ipv6 prefix and split the ipv4 range for the different networks. You can have a look at location w38b for a good example

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+      fblipke-core: 1
+      fblipke-ap01: 2
+  - vid: 50
+    role: uplink
+
+  - role: tunnel
+    ifname: ts_wg0
+    mtu: 1280
+    prefix: 10.248.3.177/32
+    wireguard_port: 51820
```
Noki, line 61: while one uplink is fine, you might want to have a second uplink for more redundancy and faster routing failovers e.g. during a firmware update of the uplink gateway.

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+      - radio: 11g_standard
+        legacy_rates: false
+        country: DE
+    ifaces:
+      - mode: ap
+        ssid: berlin.freifunk.net
+        encryption: none
+        network: dhcp
+        radio: [11a_standard, 11g_standard]
+        ifname_hint: ff
```
Noki, line 75: please remove this. Your configuration stated in the hosts section that it uses the default profile and therefore you don't need it. You only need to override the wireless_profiles if you want the APs to announce networks that are not the default. Have a look at w38b for a complex example.

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+  - vid: 40
+    role: dhcp
+    prefix: 10.248.3.128/27
+    ipv6_subprefix: 0
+    inbound_filtering: true
+    enforce_client_isolation: true
+    no_corerouter_dns_record: true
+    assignments:
+      fblipke-core: 129
+      fblipke-ap01: 130
```
Noki, line 44: You don't need this line. The ap will be reachable via the ip of the management network (vid 42).

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+height: 13
+contact_name: 'Guido L.'
+contact_nickname: 'LipkeGu'
+contacts:
+  - 'contact@fblipke.de'
+hosts:
+  - hostname: fblipke-core
+    role: corerouter
+    model: "x86_64"
+    wireless_profile: freifunk_default
```
Noki, line 16: does your core have wireless interfaces? if not remove this line as it is only required for core routers that provide wireless interfaces.

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+    low_flash: true
+  - hostname: fblipke-ap01
+    role: ap
+    model: "ubnt_nanostation_m2_xw"
+    wireless_profile: freifunk_default
```
Noki, line 21: this is the default for APs and therefore can be removed from the configuration.

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+
+ipv6_prefix: "2001:bf7:850:1500::/56"
+
+networks:
+  - vid: 20
+    role: mesh
+    name: 11s_n_2g
+    prefix: 10.248.2.200/32
+    ipv6_subprefix: -20
+    mesh_ap: fblipke-core
```
Noki, line 32: this specifies the device where the mesh interface is. in your case this should be `fblipke-ap01`

group_vars/model_ubnt_nanostation_m2_xw.yml
Noki, line 1: pleake make adding the model file a new commit with commit message `model: add nanostation_m2_xw` or similar.

locations/fblipke.yml
Noki, line 1: just to point it out, as this is a private setup: In the current version this will install all the community ssh keys present in `/group_vars/all/ssh-keys.yml` onto your router. If it is properly isolated from your network and this is fine you can keep it that way and add your own SSH key by merging it in (https://github.com/freifunk-berlin/bbb-configs/blob/cad7136daf4a9bfb4719292aabb147721e0fbabe/locations/kotti.yml#L90-L92) or if you only want it to be accessible by yourself you should override the SSH keys (https://github.com/freifunk-berlin/bbb-configs/blob/cad7136daf4a9bfb4719292aabb147721e0fbabe/locations/w38b.yml#L249-L252)

locations/fblipke.yml
```diff
@@ -0,0 +1,75 @@
...
+    mesh_radio: 11g_standard
+    mesh_iface: mesh
+  - vid: 40
+    role: dhcp
+    prefix: 10.248.3.128/27
+    ipv6_subprefix: 0
+    inbound_filtering: true
+    enforce_client_isolation: true
+    no_corerouter_dns_record: true
```
Noki, line 41: this is the main dhcp_network for the router, therefore this setting should not be added.

Noki (changes requested):

group_vars/model_ubnt_nanostation_m2_xw.yml
```diff
@@ -0,0 +1,29 @@
+---
+override_target: "ubnt_nanostation-m"
+target: ath79/tiny
```
Noki, line 3: I think this should be ath79/generic (https://firmware-selector.openwrt.org/?version=23.05.4&target=ath79%2Fgeneric&id=ubnt_nanostation-m-xw)

PR #936, opened 2024-08-13
kiezladen154: fix config typo

Noki:


Noki:
needs to be flashed, however the location is currently not reachable

PR #938, opened 2024-08-13
maintenance: suedblock, kub, dragonkiez-adlerhalle

pktpls:
- dragonkiez-adlerhalle is the last Unifi AC Mesh to be upgraded to 23.05
- kub's ap1 Fritzbox has broken wifi, replace with Cudy X6 (appointment this week)
- suedblock's Fritzbox has some issue which prevents upgrades, replace with Cudy X6 (appointment this week)


pktpls:
It's already upgraded an I'm removing the reminder comment :+1:

PR #950, opened 2024-08-27
packages: no debug packages with low_mem

Noki:
these packages add to the size of the image that will be uploaded to /tmp/ and should therefore be removed for low_mem.


pktpls:
Sounds good

PR #953, opened 2024-08-27
Fix custom port config

FFHener:


Akira25:
Looks good to me so far. What do you want to achieve with these changes?

noxilixon (approved):

PR #956, opened 2024-08-28
model: Nanostation Loco XM (1 port) support for both 2.4 and 5 GHz

Noki:
We have a bunch of Nanostation M5/M2 Loco XM 1 with only 1 ethernet port. The installation section for these devices states that they will be bricked when installing the latest Nanostation firmware and instead the Bullet M firmware should be used (https://openwrt.org/toh/ubiquiti/nanostationm5#installation). This PR creates a model file for a 2.4 GHz and 5 GHz version that can be used and works with the devices we have so they can be deployed in the field.


pktpls:
Looks good :+1:

PR #960, opened 2024-08-30
ak36-gw: add v6 uplink

nicolasberens:
Add a v6 address to the ak-gw.
The router will announce a default route and accept the announcement of the /44s but won't propagate them to the IXes


pktpls:
:+1:

PR #963, opened 2024-09-07
draft: add c-base

nicolasberens:


locations/c-base.yml
```diff
@@ -0,0 +1,96 @@
+---
+location: c-base
+location_nice: "Rungestraße 20, 10179 Berlin"
+latitude: 52.513451
```
noxilixon, line 4: Coordinates seem to be off. The location was shown in the water: https://hopglass.berlin.freifunk.net/#!v:m;n:c-base-core.olsr

Noki: should be set to 52.512865,13.42017

Akira25: resolved by [d109212](https://github.com/freifunk-berlin/bbb-configs/pull/963/commits/d109212878319ddab1de45ff8453fa6c8383fbea)

Issue #964, opened 2024-09-11
Route IPv4 via Babel


spolack:
In order to get rid of olsrd to replace it with babel we should find a way, to not interrupt Routing for nodes which are not yet upgraded to avoid taking them down.

Proposed procedure:
1. Deploy Routing Changes on all core routers running babe - All which are in bbb-configs.
 
    - Add new Routing Tables to avoid babel routes compete with olsrd, which could lead to routing loops. We have to ensure that the routes get installed only in the appropriate tables, never in the default/main.
        ```
        babel-default # Contains default routes from babel
        babel-ff # Contains freifunk routes (hna) from babel
        olsr-default # Contains default routes from olsr
        olsr-ff #  Contains freifunk routes (hna) from olsr
        ```
    - Setup routing policy in following order:
        ```
        babel-ff -> olsrd-ff -> babel-default -> olsrd-default
        ```

    With these changes in place we have ensured that babeld and olsrd routes never compete and babel is prefered whenever a route is there, but if not it is routed using olsr. 
    (change is prepared in https://github.com/freifunk-berlin/bbb-configs/commit/492b13f68b772386577d5cff072a30f875f9da77)

2. Enforce similar routing changes to gateways, but keep babel-default and olsr-default out of the routing policy, because they will never be used
3. Announce default route from gateway into babel. From this moment on all traffic going towards internet from BBB is using babels routes
4. Announce HNA from all nodes into babel
5. olsrd is unused from this moment on, but still in place to carry routes to non-babel islands connected to BBB

spolack:
In order to clean up the mess of having many routing daemons i ran a PoC to verify if bird2 can be used as drop-in replacement replacing babel entirely.

spolack:
Seems the test results were promising.

Lets deploy that on our main hub sites across the backbone.

- [ ] segen-core
- [ ] sama-core
- [ ] magda-core
- [ ] spitta13-core
- [ ] liese21-core
- [ ] q216-core
- [ ] zwingli-core
- [ ] vaterhaus-core
- [ ] emma-core
- [ ] rhnk-core
- [ ] klunker-core
- [ ] philmel-core
- [ ] rhxb-core
- [ ] simeon-core
- [ ] flughafen-core
- [ ] teufelsberg-core
- [ ] mela-core
- [ ] chris-core

Once this is done we can start advertising the IPv4 default route via bird!! :)

PR #965, opened 2024-09-12
Implement IPv4 Route Policy to prefer babel routes

spolack:
Implement following changes which are preparing the migration towards using babel also for IPv4 - #964

- Add new Routing Tables to avoid babel routes compete with olsrd, which could lead to routing loops. We have to ensure that the routes get installed only in the appropriate tables, never in the default/main.
    ```
    babel-default # Contains default routes from babel
    babel-ff # Contains freifunk routes (hna) from babel
    olsr-default # Contains default routes from olsr
    olsr-ff #  Contains freifunk routes (hna) from olsr
    ```
- Setup routing policy in following order:
    ```
    babel-ff -> olsrd-ff -> babel-default -> olsrd-default
    ```

- Ensure that olsrd and babeld install IPv4 routes only in the appropriate tables. Unfortunately babeld is not able to install routes in different policies, therefore routes are installed in "intermediate" table `babel-src`, from which they are leaked conditionally into `babel-ff` and `babel-default` using bird2. In addition it has to be ensured that the babel-src table is also used to lookup IPv6 targets.
 
With these changes in place we have ensured that babeld and olsrd routes never compete and babel is prefered whenever a route is there, but if not it is routed using olsr.

### DIFF
--- a/group_vars/role_corerouter/imageprofile.yml
+++ b/group_vars/role_corerouter/imageprofile.yml
@@ -1,5 +1,7 @@
 ---
 role_corerouter__packages__to_merge:
+  - bird2
+  - bird2c
   - babeld
   - luci-app-babeld
   - collectd-mod-dhcpleases
--- a/roles/cfg_openwrt/files/common/iproute2/rt_tables
+++ b/roles/cfg_openwrt/files/common/iproute2/rt_tables
@@ -1,0 +1,20 @@
+#
+# reserved values
+#
+128	prelocal
+255	local
+254	main
+253	default
+
+0	unspec
+#
+# local
+#
+#1	inr.ruhep
+
+10	babel-ff
+11	babel-default
+12      babel-src
+20      olsr-ff
+21	olsr-default
+
--- a/roles/cfg_openwrt/files/corerouter/babeld.conf
+++ b/roles/cfg_openwrt/files/corerouter/babeld.conf
@@ -1,0 +1,1 @@
+export-table 12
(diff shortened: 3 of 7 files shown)


Noki (approved):
Installed it on w38b-core an also my test router noki-core, so far without any issues. Seems to work exactly the same as before.

spolack:
@Noki can you confirm the route policy and routing tables are filled accordingly to the scheme?

```
ip rule
ip -6 rule

ip route show table babel-ff
ip route show table olsr-ff
ip route show table babel-default
ip route show table olsr-default

```

Noki:
> @Noki can you confirm the route policy and routing tables are filled accordingly to the scheme?
> 
> ```
> ip rule
> ip -6 rule
> 
> ip route show table babel-ff
> ip route show table olsr-ff
> ip route show table babel-default
> ip route show table olsr-default
> ```

Yes, from my understanding that is the case. Looks good to me. Looking forward for a first babel-default route for IPv4 ;-)

spolack:
I'd propose to depoloy that change to a few high traffic nodes like
- Sama
- Zwingli
- Magda 
- RHNK

in order to check for sideeffects we might have overseen yet at our stub sites.

PolynomialDivision (approved):

PolynomialDivision:
If you look at the babel neighbours, we see that there is no ipv4 neighbours but ipv4 routes. Is that intended?

Noki:
> If you look at the babel neighbours, we see that there is no ipv4 neighbours but ipv4 routes. Is that intended?

@PolynomialDivision you might have spotted the routes from the f2a nodes etc. setup by patrick. They were in there before...

PolynomialDivision:
@spolack what about announcing the default ipv4 via babeld?

PR #970, opened 2024-09-17
tempelwg: add new location

PolynomialDivision:


locations/tempelwg.yml
```diff
@@ -0,0 +1,167 @@
...
+        radio: 11a_standard
+        ifname_hint: pr
+
+      - mode: ap
+        ssid: o2-WLAN68
+        encryption: sae-mixed
+        key: 'file:/root/wifi_pass'
+        network: prdhcp
+        radio: 11g_standard
+        ifname_hint: pr
```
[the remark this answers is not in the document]
PolynomialDivision: thanks for spotting. just copied it from another location

locations/tempelwg.yml
```diff
@@ -0,0 +1,159 @@
...
+    wireless_profile: tempelwg
+
+ipv6_prefix: "2001:bf7:810:1500::/56"
+
+dhcp_no_ping: false
+
+# Reserved Prefixes:
+# 10.248.17.0/26, 2001:bf7:810:1500::/56
+# MGMT: 10.248.17.0/28
+# MESH: 10.248.17.16/26
```
Noki, line 27: should be /28

locations/tempelwg.yml
```diff
@@ -0,0 +1,159 @@
...
+latitude: 52.465551
+longitude: 13.38598
+contacts:
+  - "@nick:matrix.riotcat.org"
+
+hosts:
+  - hostname: tempelwg-core
+    role: corerouter
+    model: "dlink_covr-x1860-a1"
+    wireless_profile: tempelwg
```
Noki, line 13: missing mac-override. mac could be read with this command: https://github.com/freifunk-berlin/bbb-configs/blob/main/group_vars/model_dlink_covr_x1860_a1.yml#L12

locations/tempelwg.yml
```diff
@@ -0,0 +1,159 @@
...
+    mtu: 1280
+    prefix: 10.248.17.19/32
+    wireguard_port: 51821
+
+location__channel_assignments_11g_standard__to_merge:
+  tempelwg-core: 13-20
+  tempelwg-ap-tini: 1-20
+
+location__channel_assignments_11a_standard__to_merge:
+  tempelwg-core: 36-20
```
Noki, line 96: Why only 20 MHz channel width? Intentional even with access to a private network?

locations/tempelwg.yml
PolynomialDivision: But we have no mac override fur eth1? However similar locations with the same device also only have eth0 mac override...

Noki: on my router it looks like this:

image

My understanding is that both ports are connected to a switch (https://github.com/openwrt/openwrt/blob/d0a2b3f824c5e283caf31ceea49abf3022e7dc51/target/linux/ramips/dts/mt7621_dlink_covr-x1860-a1.dts#L191-L203) and therefore there is only eth0, correct me if I'm wrong.

The real question for me is why isn't the mac properly set. From my undestanding and looking at the DAP X1860 the dts file might just miss the part where the mac-address is read: https://github.com/openwrt/openwrt/blob/d0a2b3f824c5e283caf31ceea49abf3022e7dc51/target/linux/ramips/dts/mt7621_dlink_dap-x1860-a1.dts#L203-L204

It might be a good idea to actually just fix the board definitions for all devices where we currently use mac_overrides.

PolynomialDivision: Yeah you are right. But I also have an eth1 interface if you look into `ip l`. ^^

PolynomialDivision: Did you try without doing a mac override? For me the dts looks at first glance okay? What do you want to add to the dts? Maybe you are right but I need a closer look.

PR #974, opened 2024-09-22
workflows: perform mac-override-check only on changed files

Noki:


PolynomialDivision:
Works perfectly fine. :)

PR #981, opened 2024-09-24
tunspace: stability fixes, direct mode, ephemeral keys

pktpls:
Context: https://github.com/freifunk-berlin/falter-packages/pull/438

Tunnel client hosts need this PR because tunspace now expects the `uplink_mode` setting.


Noki (approved):
tested and works just fine

Issue #985, opened 2024-09-28
add check for "-" in interface names


Noki:
Interface names should not contain a "-", otherwise the generated config will be broken. Create a check for that.

PR #987, opened 2024-09-29
workflows: sanity check for interface names

Noki:
closes https://github.com/freifunk-berlin/bbb-configs/issues/985


Noki:
@FFHener we already have length checks in the templates. ;-)

PR #998, opened 2024-10-06
gruni73: document PoE watchdog

Noki:


Noki:
@robertfoss I activated the PoE watchdog (10m) for the 3 nf devices as they are constantly showing up in UISP as in outage. I think this is the best solution for now, what do you think?

PR #1005, opened 2024-10-14
forcki: added again

### LINKED ISSUE #782: forcki: readd location

FFHener:
fixes #782
Added a second tunnel and cleaned the config up a bit

- [ ] Flash location


Noki:
I think we should also add a 2 GHz mesh to increase chances of meshing. Looks good besides that.

Issue #66, opened 2021-09-25
meshing on core router doesnt require setting up bridge and vlan


spolack:
That could be simplified

/etc/config/wireless
```
config wifi-iface 'radio0_if2'
	option device 'radio0'
	option ifname 'wlan5-mesh'
	option mode 'mesh'
	option mesh_id 'Mesh-Freifunk-Berlin'
	option mesh_fwding '0'
	option mcast_rate '12000'
	option network 'mesh_core'
```

/etc/config/network
```
config interface 'mesh_core'
	option ifname 'eth0.20'
	option proto 'static'
	option ipaddr '10.31.114.0/32'
        option type 'bridge'
	option ip6addr '2001:bf7:830:26ff::1/128'
```

Would it work like that?

/etc/config/network
```
config interface 'mesh_core'
	option proto 'static'
	option ipaddr '10.31.114.0/32'
	option ip6addr '2001:bf7:830:26ff::1/128'
```

pmelange:
First of all, the following are no longer valid:
```
config interface 'mesh_core'
     option ifname 'eth0.20' # ifname is not longer a valid option for an interface
     option type 'bridge' # type is no longer a valid option for an interface
```

but having the following works with my tests and is the way the standard firmware has always been working:
```
config wifi-iface 'wlan0_mesh_2'
	option device 'radio0'
	option ifname 'wlan0-mesh-2'
	option network 'wireless0'
	option encryption 'none'
	option mcast_rate '12000'
	option mesh_fwding '0'
	option mode 'mesh'
	option mesh_id 'Mesh-Freifunk-Berlin'
```
```
config interface 'wireless0'
	option netmask '255.255.255.255'
	option proto 'static'
	option ipaddr '10.31.34.68'
	option ip6assign '64'
```

spolack:
Its implemented. Result is visible in huette-core config

```
config interface 'loopback'
	option device 'lo'
	option proto 'static'
	option ipaddr '127.0.0.1'
	option netmask '255.0.0.0'
	option dns '46.182.19.48 80.67.169.40 194.150.168.168 2001:4ce8::53 2001:910:800::12'
	option ip6addr '2001:bf7:830:2600::/56'



config interface 'mesh_core'
	option proto 'static'
	option ipaddr '10.31.114.0/32'
	option ip6addr '2001:bf7:830:26ff::1/128'


config interface 'mesh_huette'
	option device 'eth0.21'
	option proto 'static'
[...]

PR #1009, opened 2024-10-15
nftables: omit reflection filter on corerouters

pktpls:
(if and endif are the only changes, the rest is indentation fixes.)

Intermittently breaks routed IP traffic to mesh neighbours if the mesh interface is using the underlying (wifi) interface directly, instead of a bridge or VLAN.

The filter is supposed to learn our own mac addresses from outgoing traffic and then reject incoming reflected packets with our own src address. If we don't filter these, logread gets cluttered with informational messages saying "received packet with our own address".

However, if a mesh interface uses the underlying interface directly (instead of wrapping it in a bridge or VLAN), we somehow end up blocking our mesh neighbours. This is the case e.g. on corerouters that mesh on their own, with integrated antennas.

For now let's "fix" this by not using the reflection filter on corerouters anymore.

### DIFF
--- a/roles/cfg_openwrt/templates/common/nftables.conf.j2
+++ b/roles/cfg_openwrt/templates/common/nftables.conf.j2
@@ -60,20 +60,71 @@ table bridge client_isolation {
   {% endif %}
 {% endfor %}
 
-{% for network in networks | selectattr('role', 'equalto', 'mesh') | selectattr('name','in', network_ifname_map|map(attribute='network')) %}
-  {% set wifi_if = network_ifname_map | selectattr('network', 'equalto', network['name']) | map(attribute='ifname') | first %}
-  {% set set_localrouter = 'localrouter_' + network['name'] %}
-  {% if loop.first %}
+{#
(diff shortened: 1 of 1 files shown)


pktpls:
Ok I fixed this, and also added a bunch of information as a template comment

Noki:
Thanks for the explanation! As I understand it, we could also introduce bridges for the WLAN interfaces on the router in setups with core routers and additional APs and then filter them cleanly in the same way as with the APs. This would prevent us from having log spam. Is there anything that speaks against doing this or even always using bridges for the WiFi interfaces? Is the performance impact not negligible here, especially considering that we do the same for all other mesh interfaces?

pktpls:
Yes we did that in the past. Context https://github.com/freifunk-berlin/bbb-configs/issues/66

Noki:
Ok, so let's go back to bridges and document the reasoning behind it as comments in code?!?

pktpls:
You discuss that with Perry, this PR is getting merged

PR #1010, opened 2024-10-16
mesh: fixed / added missing 60 GHz mesh metrics

Noki:
A lot of 60 GHz links miss mesh metrics and therefore use the default of 512. In addition some links seem to use mesh_metrics that are higher than required. This PR sets the mesh_metric to 128 for 60 GHz links.

The following locations needs to be flashed for this to become active.

- [ ] ak36
- [ ] gub37
- [ ] ilr
- [ ] kiehlufer
- [ ] klunker
- [ ] ohlauer
- [ ] radbahn
- [ ] rhnk
- [ ] sav
- [ ] scharni
- [ ] teufelsberg
- [ ] vaterhaus
- [ ] wilgu10

Afterwards `metrix.sh` should be executed at the following gateways:

- [ ] ak36
- [ ] ohlauer
- [ ] saarbruecker


Noki:
I flashed all the locations relevant for the routing, the other locations can be flashed later on.

PR #1011, opened 2024-10-19
Fix mesh metrics

Noki:
Looks like metric was renamed to mesh_metric at some point to be identical accross gateways and corerouters, without properly adjusting all config files:

https://github.com/freifunk-berlin/bbb-configs/blob/28dbaf6b4ea99a6dbd392ac7fc575e82cbcd91fa/roles/cfg_openwrt/templates/gateway/config/babeld.j2#L22

https://github.com/freifunk-berlin/bbb-configs/blob/28dbaf6b4ea99a6dbd392ac7fc575e82cbcd91fa/roles/cfg_openwrt/templates/corerouter/config/babeld.j2#L21

This PR fixes this so we generate proper babeld config files. In addition this fixes the mesh metric at w38b that was still to high.

### DIFF
--- a/locations/ak36.yml
+++ b/locations/ak36.yml
@@ -80,21 +80,21 @@ mesh_links:
     ifname: eth1.10
     ipv4: 10.31.130.160/32
     ipv6: 2001:bf7:750:4001::1/128
-    metric: 1024
+    mesh_metric: 1024
     ptp: true
 
   - name: mesh_flughafen
     ifname: eth1.11
     ipv4: 10.31.130.161/32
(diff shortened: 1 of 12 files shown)


robertfoss:
LGTM

pktpls:
:grimacing:

pmelange:
> @pmelange what was the reason to set the mesh-metric to rhxb lower than the local mesh? As the variable wasnt set correctly so far, did you experience any issues that would be resolved by worsen this link?

No, I believe I just copied what was already done for dragonkiez-kiezraum for the other dragonkiez locations.

PR #1016, opened 2024-10-21
treewide: remove mt76 debugging

pktpls:
(Except on radbahn-w-nf and radbahn-o-nf, which I'll monitor for a while longer.)


Noki (approved):
looks good to me

PR #1020, opened 2024-10-22
 Revert more whitespace changes from PR #1013

robertfoss:


robertfoss:
I'm only seeing intended changes from #1013  now, so I think we're good to merge this.

PR #1021, opened 2024-10-22
imageprofile: return to MbedTLS

pktpls:
There was a time when MbedTLS was already the default in OpenWrt, but didn't yet support everything that we needed.

It does now. I have been testing it on a few busy locations for a few months.


Noki (approved):
tested and works for me

Issue #1012, opened 2024-10-20
Monitoring: Improve labels for snmp-devices


FFHener:
We currently use the OID `.1.3.6.1.2.1.31.1.1.1.18` to get the labels of the ports. Unfortunantly I discovered that these labels exist for EdgeSwitch 10XP running 1.9.0 (latest), but not for EdgeSwitch 8XP running 2.2.1 (latest).

The only thing for labels I found at the first glance would be
```
IF-MIB::ifDescr.1 = STRING: Port 1
IF-MIB::ifDescr.2 = STRING: Port 2
[...]
IF-MIB::ifDescr.8 = STRING: Port 8
IF-MIB::ifDescr.11 = STRING: Management
```
which still might be an upgrade over the current thing defaulting to the Instance ID which would be `1` to `11`. 

I thing one path going forward could be to have one profile per device type which would require defining the explizit type in the config. Would be some work, but a nice preparation for furhter features like automated config for non-owrt-devices.

Another path could be to automate the snmp-Monitoring on the Grafana/Prometheus-Side - which would be an even bigger chunk of work. The first solution would enable us at least to go in smaller steps. Im happy for comments and opnions on this

Noki:
I think the way to go is to describe every device in a model file in group_vars and move everything device related there (SNMP / DFS-Reset) and have every device described as a host. Maybe we can design the group vars file it in a way to use a template for settings that can be described for a set of devices (SNMP for airOS 8, switchOS, ...) but even having explicit settings per device would be fine in my opinion. We might even be able to override only a small set of the settings from a template or add more specific settings.

Issue #751, opened 2024-01-30
RFC: Changes to the configuration format to further standardize it


Noki:
I talked with @FFHener about ways to further standardize the configuration format. We came up with the following ideas:

- Add model files for non OpenWRT devices (ubiquiti antennas, switches, webcams, ...)
- Add all non OpenWRT devices as hosts alongside the OpenWRT devices
- Add configurations values to individual non OpenWRT hosts (e.g. height+altitude, alignment, ssid, ...)

Doing so will allow us to move some parts of the configuration into the new model files:

- SNMP profiles
- airOS DFS reset (default configuration, with configuration overrides for each host)

In addition this will allow us to:

- derive a proper documentation for each location that could be used to update the wiki / website
- generate configuration files for non OpenWRT devices when building the location

Here are examples of how a model file for a non OpenWRT device could look like:

```
brand_nice: Ubiquiti
model_nice: Wave Long-Range

wireless_devices:
  - name: 11ad_standard
    band: 60g
  - name: 11a_standard
    band: 5g
    htmode_prefix: HE

snmp_profile: af60
```

```
brand_nice: Ubiquiti
model_nice: Nanostation 5AC

wireless_devices:
  - name: 11a_standard
    band: 5g
    htmode_prefix: HE

snmp_profile: airos_8

airos_dfs_reset:
  - username: "ubnt"
    password: "file:/root/pwd"
    daytime_limit: "2-7"
```

The host configuration for such an device could look like this:

```
hosts:
  - hostname: q216-tierpark
    role: ptp-ap
    ptp: true
    model: "ubnt_nanobeam_5ac-gen2"
    height: 15
    altitude: 54
    orientation: ost
    ssid: "q216-tierpark" #to override default ssid (freifunk-{location}-{orientation}-{frequency} / freifunk-{location}-{orientation})
    channel: 128
    channel-width: 40
    username: "root" # to override the default ("ubnt")
    password: "file:/root/tierpark-pwd.txt" # to override the default ("file:/root/pwd.txt)
    daytime_limit: "4-5"  # to override the default ("2-7")
```

We want to discuss this at the community day. Please let us know about any thoughs you have on this.

FFHener:
I added some comments in the proposed code. Did you check with the config files from the properitery devices that we need the wireless_devices in this way?

Noki:
Well, we need some way to know the band and we also need to know if it is a dual band device, so probably a good idea to just keep the syntax we already have and build on that.

[...]

PR #1028, opened 2024-10-24
Enable non-openwrt devices as hosts

FFHener:
This is a first step into implementing #751. It currently only supports hosts that doesnt need `airos_dfs_reset`. The changes wont break the current way of defining everything. This way we can make a smooth transition.

Im not to shure what could be the best way to deal with locations that doesnt use an own mgmt-network.

Im not super experienced in writing jinja templates. Feel free to suggest better ways.

Merging this PR will enable us to move forward with #1012

### DIFF
--- a/roles/cfg_openwrt/templates/common/collectd/conf.d/bbb-snmp.conf.j2
+++ b/roles/cfg_openwrt/templates/common/collectd/conf.d/bbb-snmp.conf.j2
@@ -1,9 +1,9 @@
 #jinja2: trim_blocks: "true", lstrip_blocks: "true"
 
-{% set profiles = snmp_devices | default([]) | map(attribute='os') | list | unique %}
+{% set profiles = snmp_devices | default([]) | map(attribute='os') | reject('search', 'openwrt') | list | unique %}
 {# collect profiles from new way of writing hosts as well and append them #}
 {% for host in groups[group_names[0]]%}
-  {% if hostvars[host]["os"] is defined %}
+  {% if hostvars[host]["os"] != "openwrt" %}
     {% do profiles.append(hostvars[host]["os"]) %}
   {% endif %}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


roles/cfg_openwrt/templates/common/collectd/conf.d/bbb-snmp.conf.j2
```diff
@@ -1,6 +1,12 @@
 #jinja2: trim_blocks: "true", lstrip_blocks: "true"
 
 {% set profiles = snmp_devices | default([]) | map(attribute='snmp_profile') | list | unique %}
+{# collect profiles from new way of writing hosts as well and append them #}
+{% for host in groups["location_q216"]%}
```
Noki, line 5: this is location specific and needs to be changed

Noki:
We should have one or two example of devices that can run more than one OS e.g. the stock OS and OpenWRT or SwOS and RouterOS. How do we name model files in this case?

Ideas:

/group_vars/models/[os]/[name].yml
/group_vars/model_[os]_[name].yml

Currently I can't think of a good way of doing this without also touching all location files, but maybe we can also implement a fallback.

Noki:
For the snmp_devices we can and should continue to use snmp_profile and simply set it as "os" in order to be able to use it in the j2 template, because if we start changing something in the snmp_devices, we can also migrate everything at the same time, but that makes the PR extensive and therefore difficult to handle.

Suggestion:

https://github.com/freifunk-berlin/bbb-configs/blob/ba885ce098766ff1ea8dc1c4c0624c5121685897/roles/cfg_openwrt/templates/common/collectd/conf.d/bbb-snmp.conf.j2#L4

We insert a step here that runs through the profiles in a loop and ensures that os is set and filled with the value of snmp_profile so that the old notation is compatible. The changes to the locations are then optional.

Let me know if you have any questions regarding this proposal.

roles/cfg_openwrt/tasks/main.yml
```diff
@@ -12,6 +12,7 @@
     apply:
       tags: always
   tags: always
+  when: os == "openwrt" | default ("openwrt")
```
spolack, line 15: I think there are much better ways to archieve that. Lets have a workshop or chat on that.

Issue #995, opened 2024-10-03
RFC: Replace babel with bird2


spolack:
In order to shrink our zoo of routing daemons i'd propose to replace babeld with bird2. I'm currently running a PoC configuration at wilgu10-core. This also allows us to get rid of the "hack" trying to let bird workaround missing features (kernel export filter) in babeld. Other communities like freifunk franken also switched to bird because it's performing better. Even Jech thinks that birds implemenation is superior.

PoC Config running at wilgu10-core. From what i can tell it seems to work and is interoperable with babeld.

```
root@wilgu10-core:~# cat /etc/bird.conf 
log syslog all;
debug protocols all;

ipv6 sadr table v6_main;

ipv4 table v4_babel_src;
ipv4 table v4_babel_ff;
ipv4 table v4_babel_default;

protocol device {
};

protocol direct {
	interface "lo";
	ipv6 sadr {
		table v6_main;
	};
}



protocol babel {
	randomize router id yes;
        ipv6 sadr {
                table v6_main;
                import where (net.len = 56) || (net = ::/0 from 2001:bf7::/32 );
                export filter {
			if net = 2001:bf7:860:900::/56 from ::/0 then accept;
			if source = RTS_BABEL then accept;
			reject;
		};
        };
        ipv4 {                                          
                table v4_babel_src;                                   
                import all;      
                export where source = RTS_BABEL;
        };
	
	interface "switch0.10" {
		type wireless;
		rxcost 128;
	};
        interface "switch0.11" {
                type wireless;
                rxcost 1024;
        };
        interface "switch0.20" {
                type wireless;
                rxcost 2048;
        };
}

protocol kernel {       
        ipv6 sadr {                  
                table v6_main;
                export all;
        };       
}   

protocol kernel {
        kernel table 12;
	ipv4 {
                table v4_babel_src;
		export all;
	};
}

protocol kernel { 
        kernel table 10;
	ipv4 {
                table v4_babel_ff;
		export all;
	};
}

protocol kernel { 
        kernel table 11;
        ipv4 {             
        	table v4_babel_default;
                export all;
        };                 
}

protocol pipe {
	table v4_babel_src;
	peer table v4_babel_ff;
	export where net != 0.0.0.0/0;
}

protocol pipe {
        table v4_babel_src;
        peer table v4_babel_default;
        export where net = 0.0.0.0/0;
```

Noki:
[...]

Issue #969, opened 2024-09-17
Document and enforce mesh_metric


spolack:
We want to start routing IPv4 via Babel soonish. Lets make sure that the metrics are in the desired state to have a deterministic routing experience :)

- [ ] Agree on metrics concept
- [ ] Document metrics
- [ ] Ensure metrics are set over most criticial PoPs

spolack:
RFC 0.1 
| Metric | \>=Speed | Remarks |
| --- | --- | --- |
| 128-511 | &nbsp;5 Gbit/s | For future use |
| 512-4095 | 1 Gbit/s | 60Ghz |
| 4096-8191 | 100 Mbit/s | 5GHz |
| 8192-16383 | 50 Mbit/s | 5GHz PtMP |
| 16384-32767 | 10Mbit/s | 2GHz / Default) |

Metric: The first value should be tried first. In case of suboptimal routing the the range can be used as wiggle-room for individual traffic engineering.

By applying this scheme we would prefer:

- 4 5GBit/s links over single 1Gbit Link
- 8 1Gbit/s links over single 100Mbit/s link
- ...

Noki:
I think we should be a bit more granular especially in the area between 1 Gbit/s and 100 Mbit/s so we can set values that match the speed of the connections more closely e.g. by measuring it with iperf3 or looking at the capacities reported by UISP. In my opinion this would help in areas where there are multiple 5 GHz paths with a lot of variance in bandwidth. I know that this was meant with `wiggle room` but I would prefer a more detailed table as it leaves less room for interpretation. In addition I think it is also a good idea to have 2.5 Gbit/s as we already have it with the airFiber 60 Xtreme-Range (ak36<->teufelsberg), eventhough the cable connection/core-router is currently the limit there.

My proposal would be as follows:

| **Metric**      | **Max Speed**   | **Min Speed >**   | **Remarks**                           |
|-----------------|-----------------|-----------------|---------------------------------------|
| 64 - 127        | 10 Gbit/s       | 5 Gbit/s        | Fiber                                 |
| 128 - 255       | 5 Gbit/s        | 2.5 Gbit/s      | Fiber                                 |
| 256 - 511       | 2.5 Gbit/s      | 1 Gbit/s        | 60 GHz wireless/ethernet              |
| 512 - 1023      | 1 Gbit/s        | 500 Mbit/s      | 60 GHz wireless/ethernet              |
[...]

PR #1038, opened 2024-10-29
Ready for testing: Babel via bird 

spolack:
Replacing babeld and switch to bird2 standalone. Only core routers are covered for now, that might change.

#995


roles/cfg_openwrt/templates/corerouter/bird.conf.j2
```diff
@@ -0,0 +1,105 @@
...
+	interface "{{ libnetwork.getPortIfname(nw) }}" {
+		type wireless;
+		rxcost {{ nw.get('mesh_metric', default_mesh_metric) }};
+	};
+{% endfor %}
+	# Tunnel interfaces provided by tunspace
+{% for nw in networks | selectattr('role', 'equalto', 'tunnel') %}
+	interface "{{ nw.get('ifname') }}" {
+		type wireless;
+		rxcost {{ nw.get('mesh_metric', mesh_metric_tunnel_in) }};
```
PolynomialDivision, line 52: are u sure tunnels are included? does not look like I get a route from bird via wireguard

PolynomialDivision (approved):

Noki (approved):
works at w38b

group_vars/all/general.yml
```diff
@@ -42,3 +42,16 @@ all_sysctl__to_merge:
...
+freifunk_global_prefix: 2001:bf7::/32
+
+# Default mesh metric in inbound direction (rxcost) for normal mesh links
+mesh_metric_default_in: 512
+
+# Default mesh metric in inbound direction (rxcost) for tunnels
+mesh_metric_tunnel_in: 512
+
+# Default mesh metric in inbound direction (rxcost) for adhoc like interfaces
+mesh_metric_adhoc_in: 2028
```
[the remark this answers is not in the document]
spolack: fixed, thx

roles/cfg_openwrt/templates/corerouter/bird.conf.j2
```diff
@@ -0,0 +1,104 @@
...
+		};
+	};
+	ipv4 {
+		import all;
+		export where source = RTS_BABEL; # For now only advertise routes learned from babel
+	};
+
+	# Mesh interfaces
+{% for nw in networks | selectattr('role', 'equalto', 'mesh') %}
+  {% set default_mesh_metric = mesh_metric_adhoc_in if 'mesh_ap' else mesh_metric_default_in %}
```
Noki, line 41: How about taking this a bit further by looking at the value of `mesh_radio` in addition and give `11g_standard` a slightly worse value than `11a_standard` to prefer 5 GHz links in dense areas like k12, hdk, weidenbaum, kiehlufer. By doing so we could get rid of quite some manually defined mesh_metrics in the configs that were only introduced to make 2 GHz worse. What do you think?

group_vars/all/general.yml
```diff
@@ -42,3 +42,16 @@ all_sysctl__to_merge:
...
+
+
+## Routing Section:
+freifunk_global_prefix: 2001:bf7::/32
+
+# Default mesh metric in inbound direction (rxcost) for normal mesh links
+mesh_metric_default_in: 512
+
+# Default mesh metric in inbound direction (rxcost) for tunnels
+mesh_metric_tunnel_in: 512
```
Noki, line 54: I think in relation to the other metrics where we set ad_hoc to 2048 this default metric is way to low. My understanding was that for nodes with a connection to the mesh we want to have wireless routing preferred and treat tunnels as a failover. With this my default route via the tunnels has a metric of 1024 and the lowest metric via one of the wireless interfaces is 2560 (best case via wireless would be 2048 (5 GHz adhoc) + 128 (60 GHz) + 128 (60 GHz) = 2304).

I think we really need to decide about https://github.com/freifunk-berlin/bbb-configs/issues/969 and should probably base the default on a 50 MBit connection.

Noki:
Works for me with the changes made but I had to change mesh metrics to get the desired routing.

roles/cfg_openwrt/templates/corerouter/bird.conf.j2
```diff
@@ -0,0 +1,104 @@
...
+	interface "{{ libnetwork.getIfname(nw) }}" {
+		type wireless;
+		rxcost {{ nw.get('mesh_metric', default_mesh_metric) }};
+	};
+{% endfor %}
+	# Tunnel interfaces provided by tunspace
+{% for nw in networks | selectattr('role', 'equalto', 'tunnel') %}
+	interface "{{ nw.get('ifname') }}" {
+		type wireless;
```
Noki, line 50: Should be `tunnel` instead of `wireless` to speed up the process of dropping routes e.g. if one of multiple tunnel servers becomes unreachable.

spolack: https://bird.network.cz/?get_doc&v=20&f=bird-6.html#ss6.2
> 
> type wired|wireless|tunnel
> 
> >     This option specifies the interface type: Wired, wireless or tunnel. On wired interfaces a neighbor is considered unreachable after a small number of Hello packets are lost, as described by limit option. On wireless interfaces the ETX link quality estimation technique is used to compute the metrics of routes discovered over this interface. This technique will gradually degrade the metric of routes when packets are lost rather than the more binary up/down mechanism of wired type links. A tunnel is like a wired interface, but turns on RTT-based metrics with a default cost of 96. Default: wired. 


While i think this is definately considerable, at least to change that to wired, i would like to keep it out. Tbh, changing any fundamential routing parameters without good testing in the field is out of scope for this PR.

roles/cfg_openwrt/templates/corerouter/bird.conf.j2
spolack: Good idea, feel free to open a PR.  Happy to review!

spolack:
Awaiting testresults in conjunction with tunspace, before merge.

In the meantime i'm going to deploy that change on all BBB Hub sites, in order to make some progress here.

group_vars/all/general.yml
Noki: But from what I can tell the metric for the default route is now lower with birdc:

babel:

```
root@kirchhof-core:~# echo "dump" | nc ::1 33123 | grep 'prefix ::/0'
add route b6e7e610 prefix ::/0 from 2001:bf7::/32 installed yes id f0:83:53:53:82:be:48:d8 metric 1536 refmetric 0 via fe80::1 if ts_wg0
add route b6e7c610 prefix ::/0 from 2001:bf7::/32 installed no id f0:83:53:53:82:be:48:d8 metric 1600 refmetric 64 via fe80::1 if ts_wg1
root@kirchhof-core:~#
```

bird:

```
root@lacasahellersdorf-core:~# birdc show babel route | egrep '^::/0'
::/0 from 2001:bf7::/32                                fe80::1                   ts_wg0      1088 * 21935  55.131
root@lacasahellersdorf-core:~#
```

So looks like a change to me.

spolack:
It does only have effect on core routers, except the `network:` commits. But it should be a no-change for those.

spolack:
Merging as it seems to run stable across our network including tunspace locations.

PR #1052, opened 2024-11-04
Kub add mesh

FFHener:
~Should be reviewed from @pktpls before flashing.~ => OOB-Approval

### DIFF
--- a/locations/kub.yml
+++ b/locations/kub.yml
@@ -14,61 +14,11 @@ hosts:
     model: "avm_fritzbox-7530"
     wireless_profile: freifunk_default
 
-  # Special vlan config:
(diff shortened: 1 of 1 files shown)


locations/kub.yml
```diff
@@ -89,10 +39,34 @@ networks:
...
+    mesh_iface: mesh
+
+  # MESH - 2.4 GHz 802.11s
+  - vid: 21
+    role: mesh
+    name: mesh_2g_ap1
+    prefix: 10.31.139.18/32
+    ipv6_subprefix: -21
+    # make mesh_metric(s) for 2GHz worse than 5GHz
+    mesh_metric: 1024
```
Noki, line 61: this line should be removed as we now prefer 5 GHz over 2.4 GHz for 802.11s by default

Issue #983, opened 2024-09-25
List of management VLANs


Noki:
For the 60 GHz devices we need unique management VLANs so it does not span accross locations. Please add new locations to this list to reserve a VLAN or comment with a VLAN number below to reserve a VLAN. Admins should move VLANs from comments to this list for a better overview.

| VLAN | location | status |
| --- | --- | --- |
| 420 |	rhnk | active |
| 421 |	kiehlufer  | active |
| 422 |	rhxb | |
| 423 |	dtmb | |
| 424 |	| |
| 425 |	zwingli | |
| 426 |	sama | |
| 427 |	ak36 | |
| 428 |	flughafen | |
| 429 |	l105 | |
| 430 |	saarbruecker | |
| 431 |	agym | |
| 432 |	| |
| 433 |	gub37 | active |
| 434 |	w38b | active |
| 435 |	ilr | active |
| 436 |	scharni | active |
| 437 |	teufelsberg | active |
| 438 |	philmel | |

Feel free to fill the empty lines and to migrate existing locations over to using the new VLANs.

PR #1053, opened 2024-11-04
bht: cleanup

FFHener:
As we will swap out the Router and Switch anyway I thought I would bring the location closer to our convention. Should we add some Nearfieldstuff as well? Does it make sense there?

This PR will be deployed On-Site at 7th of November

EDIT: Changed mgmt-IP according to https://github.com/freifunk-berlin/bbb-configs/issues/983

### DIFF
--- a/locations/bht.yml
+++ b/locations/bht.yml
@@ -1,7 +1,7 @@
 ---
 
 location: bht
-location_nice: Berliner Hochschule fuer Technik
+location_nice: Berliner Hochschule fuer Technik, Luxemburger Straße 10, 13353 Berlin
 latitude: 52.544407831736
 longitude: 13.352562785148
 altitude: 88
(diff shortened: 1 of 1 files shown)


locations/bht.yml
```diff
@@ -122,94 +124,98 @@ airos_dfs_reset:
...
+    prefix: 10.230.23.137/32
+    ipv6_subprefix: -19
 
-  - vid: 104
+  - vid: 40
     role: dhcp
-    prefix: 10.230.23.152/29
+    prefix: 10.230.23.144/28
     ipv6_subprefix: 0
     untagged: true
```
Noki, line 206: let's untag a switchport instead

pmelange:
There is no need for near field at this location. And therefore we don't need a bigger dhcp network. The old 4909 was hardly ever used because of it's location. It was good to have when doing maintenance. Now with the rb750gr3 we have to make sure that people bring a network cable with them.

pmelange:
And why do we want to use 42 for the mgmt vlan? Wasn't that a problem that everywhere 42 is being used?

pmelange:
I like the rb759gr3. Keep it.

But why remove jup from the list? One of the goals on Thursday is to get jup dish back online.

Noki:
The connection to jup should stay for sure and we need to bring that one up again. In addition we should change mgmt to something in the 4xx range to prepare for 60 GHz so we don't need to touch it again if we build 60 GHz later on. Please use the first free one in the list: https://github.com/freifunk-berlin/bbb-configs/issues/983

locations/bht.yml
Noki: I think the only reason was that switch ports were full and the core had free ports where you could plug in and therefore it was untagged.

Noki (approved):
looks good to me

PR #1059, opened 2024-11-08
OpenWRT  24.10

spolack:


pktpls:
?? https://github.com/freifunk-berlin/bbb-configs/pull/1056

PR #1060, opened 2024-11-08
corerouter: only log errors for bird instead of all

FFHener:
This makes the logs useable again. If you need more logging infos you can manually edit `/etc/bird.conf` back to `all` for debugging.


spolack:
Is that still necessary? Thats why i added changed it to `debug protocol {state}`

PR #1061, opened 2024-11-08
birdc: metrics penalty for 2GHz adhoc

Noki:
this makes the metric of 2GHz mesh links worse than 5GHz mesh links by adding the default metric.

In addition to this we need to adjust metrics at various locations where 802.11s links have metrics manually set below 2048 as the new default of 2048 for 802.11s changes routing at various locations.

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/bird.conf.j2
+++ b/roles/cfg_openwrt/templates/corerouter/bird.conf.j2
@@ -40,8 +40,9 @@ protocol babel {
 	};
 
 	# Mesh interfaces
+{# metrics for 2 GHz adhoc get a penalty over 5 GHz adhoc so 5 GHz is preferred #}
 {% for nw in networks | selectattr('role', 'equalto', 'mesh') %}
-  {% set default_mesh_metric = mesh_metric_adhoc_in if 'mesh_ap' else mesh_metric_default_in %}
+  {% set default_mesh_metric = hostvars[inventory_hostname].get('mesh_metric_adhoc_' ~ nw.get('mesh_radio'), mesh_metric_default_in) %} 
 	interface "{{ libnetwork.getIfname(nw) }}" {
 		type wireless;
 		rxcost {{ nw.get('mesh_metric', default_mesh_metric) }};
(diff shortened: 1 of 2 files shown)


roles/cfg_openwrt/templates/corerouter/bird.conf.j2
```diff
@@ -40,8 +40,15 @@ protocol babel {
 	};
 
 	# Mesh interfaces
+{# metrics for 2 GHz adhoc get a penalty over 5 GHz adhoc so 5 GHz is preferred #}
 {% for nw in networks | selectattr('role', 'equalto', 'mesh') %}
-  {% set default_mesh_metric = mesh_metric_adhoc_in if 'mesh_ap' else mesh_metric_default_in %}
+  {% if nw.get('mesh_ap') and nw.get('mesh_radio') == '11a_standard' %}
+      {% set default_mesh_metric = mesh_metric_adhoc_in %}
+  {% elif nw.get('mesh_ap') and nw.get('mesh_radio') == '11g_standard' %}
+      {% set default_mesh_metric = mesh_metric_adhoc_in + mesh_metric_default_in %}
```
spolack, line 48: i'd prefer to define global default variables an use them accordingly:
mesh_metric_adhoc_11g_standard
mesh_metric_adhoc_11a_standard

-> hostvars[inventor_hostname].get('mesh_metric_adhoc_' ~ mesh_radio, mesh_metric_default_in)

spolack (changes requested):
Please review my inline comment

roles/cfg_openwrt/templates/corerouter/bird.conf.j2
Noki: I tried, but could not get it to work within a reasonable amount of time...

Noki: I implemented a version that defines the global default variables, but I did not manage to get the generalistic varible selection based on mesh_radio working.

spolack:
LGTM!

PR #1075, opened 2024-11-15
Revert "hds: remove location"

pktpls:
hds and hds-containers continue to operate for now

### DIFF
--- a/locations/hds.yml
+++ b/locations/hds.yml
@@ -1,0 +1,93 @@
+---
+
+location: hds
+location_nice: HdS Freifunk-Raum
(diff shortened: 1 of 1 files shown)


Noki:
> hds and hds-containers continue to operate for now

In UISP hds-ffraum, hdm-switch, hdm-p3 are offline. I think the part that is operational is Falter based and not managed by bbb-config and this PR does not reflect the current state.

spolack:
Yep, that is also my understanding that hds is no more bbb-configs.

pktpls:
Oops you're right

PR #1084, opened 2024-11-21
Update k9

iuljan:
This PR contains the customized configs for the hardware change at location k9.

A 60GHZ antenna is aligned to Sama
And the previous one is aligned to Zwingli

Loge and Größenwahn get core APs (Aruba AP11) via LAN.


locations/k9.yml
```diff
@@ -113,19 +74,37 @@ networks:
...
   - vid: 42
     role: mgmt
+    name: omgmt
     prefix: 10.31.9.208/28
     gateway: 1
     dns: 1
     ipv6_subprefix: 1
     assignments:
       k9-core: 1      # 10.31.9.209
       # k9-switch: 2  # 10.31.9.210
```
nailend, line 84: @spolack can the old MGMT be removed?

nailend: @Noki

Noki: Yes, it can be removed. Everything is configured to use the new one.

nailend:
replaced by #1121

PR #1085, opened 2024-11-21
ap: prevent lockout when installing firewall later

pktpls:
The AP image doesn't contain a firewall package or firewall settings. When installing the firewall later on, it loads default settings and locks us out of the device permanently.

Add a basic firewall config allowing mgmt access, just to prevent this improbable but unfortunate case.

(This happened at Teufelsberg the other week, people needed to go there to fix it.)


Noki (approved):
looks good to me

PR #1087, opened 2024-11-22
corerouter: use patched version of bird2 and add proper inbound metrics

spolack:
please test

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/bird.conf.j2
+++ b/roles/cfg_openwrt/templates/corerouter/bird.conf.j2
@@ -27,7 +27,23 @@ protocol babel {
 	ipv6 sadr {
 		table v6_main;
 		# Import only /56 location aggregates and default routes
-		import where (net.len = 56) || (net = ::/0 from {{ freifunk_global_prefix }} );
+		import filter {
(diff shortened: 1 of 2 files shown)


roles/cfg_openwrt/templates/corerouter/bird.conf.j2
```diff
@@ -45,14 +60,12 @@ protocol babel {
   {% set default_mesh_metric = hostvars[inventory_hostname].get('mesh_metric_adhoc_' ~ nw.get('mesh_radio'), mesh_metric_default_in) %}
```
Noki, line 60: the comment should be moved up and the line with `set default_mesh_metric` should be removed as it does not do anything here anymore

Noki (changes requested):
works for me even without the requested changes.

PolynomialDivision (approved):

PolynomialDivision:
oh wait, please add something to commit message

PolynomialDivision:
Please add

```
Link: https://bird.network.cz/pipermail/bird-users/2023-February/016705.html
```

Noki (approved):

PR #1104, opened 2024-12-03
Enable BGPDisco

spolack:


roles/cfg_openwrt/templates/corerouter/config/olsrd.j2
```diff
@@ -4,7 +4,7 @@ config LoadPlugin
 
 config LoadPlugin
 	option library 'olsrd_nameservice'
-	option suffix '.olsr'
```
pktpls, line 7: Is it possible that this resets the node lifetime on hopglass?

Noki: No, with my change (https://github.com/freifunk-berlin/falter-packages/commit/ef840094613e427d9e189e4a7284248d63bc2873) we strip the suffix and force it to .olsr afterwards, however it has still implications as the hostname displayed by hopglass will not work anymore. In addition we need to update the mass-update script and there is probably e.g. documentation that becomes outdated, and hardcoded links of olsr services might stop working, ...

PolynomialDivision: For me the implications are fine. I see them not as important to block the bgpdisco improvements that I see as enormous helpful.

PolynomialDivision (approved):

roles/cfg_openwrt/templates/corerouter/config/olsrd.j2
Noki: A small addition: bird seems to work fine on mela-core (arm_cortex-a7_neon-vfpv4) without bgbdisco, but with it bird is stuck in a crash loop and mela-core is only reachable via olsr. For me this is a showstopper and should be resolved before merging this.

pktpls:
LGTM, but can't merge as long as the Bird / armv7 crash isn't resolved

pktpls:
Sorry, ich hatte kurz den Überblick verloren - der Bird / armv7 Crash hat nix mit BGPdisco zutun

PolynomialDivision:
So the armv7 fix is merged, so we could also probably merge this?

pktpls:
Coolster Hack :)

PR #1111, opened 2024-12-09
hway: networking cleanups, outdoor ap, 24.10 testing

pktpls:


locations/hway.yml
```diff
@@ -99,10 +112,12 @@ networks:
...
 
 location__channel_assignments_11a_standard__to_merge:
-  hway-ap1: 36-40
+  hway-indoor: 36-40
+  hway-street: 44-40
 
 location__channel_assignments_11b_standard__to_merge:
-  hway-ap1: 13-20
+  hway-indoor: 13-20
+  hway-street: 5-20
```
Noki, line 120: why not make it the other way around and enable meshing at the street ap so the network could grow?

pktpls: Want to mesh with devices in my home every now and then, and also don't want a trash mesh connection with elsekiehl :+1: 

The street antenna is very visible, so anyone who is interested can just ring my doorbell.

PR #1115, opened 2024-12-11
saarbruecker: replace with new protectli vps6630

spolack:


locations/saarbruecker.yml
Noki, line 1: we use `mesh_metric_tunnel_in` now and `tunnel_metric: 1024` can be removed (all gateways are affected, can therefore also happen later)

Noki (approved):
looks good to me

PR #1124, opened 2024-12-15
gateways: fix olsrd->babel soft migration

spolack:
This involves few changes:
* Move IPv4 Static default route into bird, thus delete it from main network config
* Fix OLSR messing with policy routing (by setting its priorities super high)
* Remove OLSR dyngw which doesnt work with our policy routing approach and create a static default route HNA instead.

### DIFF
--- a/group_vars/role_gateway/imageprofile.yml
+++ b/group_vars/role_gateway/imageprofile.yml
@@ -7,7 +7,6 @@ role_uplink_gw__packages__to_merge:
   - collectd-mod-snmp6
   - olsrd
   - olsrd-mod-arprefresh
-  - olsrd-mod-dyn-gw
   - olsrd-mod-jsoninfo
   - olsrd-mod-nameservice
   - olsrd-mod-txtinfo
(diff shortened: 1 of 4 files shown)


PolynomialDivision:
Did you have a look at the gateways?

```
root@saarbruecker-gw:~# ip rule show
0:	from all lookup local
32766:	from all lookup main
32767:	from all lookup default
33100:	from all lookup babel-ff
33101:	from all lookup olsr-ff
33200:	from all lookup babel-default
34000:	from all lookup olsr-ff
34020:	from all lookup 224
34030:	from all lookup olsr-default
```

What is 224?

spolack:
Thats olsr messing with route polcies.
>	option RtTableTunnelPriority '34020'

PolynomialDivision:
> Thats olsr messing with route polcies.

Argh. I'm happy to remove olsr in near future completely. ;) Do we need olsr backwards compatibility on the gateways? Legacy nodes are not using the gateways as tunnels?

PR #1127, opened 2024-12-16
Dorfplatz changes (removed zwingli connection & renamed stuff)

spinat3000:
Still needs to be flashed


FFHener (approved):

FFHener (changes requested):

Noki:
@spinat3000 what about ap1 and ap2? Both seem to be offline.

PR #1137, opened 2024-12-17
ssh-key: add Zweis key

spinat3000:
please :)


spolack:
I'm opposing that change at a global level, assumingly you would rather like to configure it at the location level for a specific location.

PolynomialDivision:
Sorry, but I do not know who you are and cannot authorize access to all our locations at this time.

Noki:
I am also against it on a global level. In the past, setting up a global SSH key was usually proposed by someone who already had their key in the global list, and only for someone who had been active and well-known in the community for a long time, e.g. by regularly attending public meetings like Freifunk Stammtisch, Community Day or Hack Day.

PR #1139, opened 2024-12-18
Hdk some changes

Noki:


Noki:
Replaced by https://github.com/freifunk-berlin/bbb-configs/pull/1140 and https://github.com/freifunk-berlin/bbb-configs/pull/1151

PR #1143, opened 2024-12-19
gateways: fix and improve configurations from previous commits

PolynomialDivision:
This patch addresses issues and refines changes made in earlier commits.


spolack:
Hmm, thank you, but that also reverts also other changes. I rebased my changeset in https://github.com/freifunk-berlin/bbb-configs/pull/1144 and added a few other optimizations.

PR #1146, opened 2024-12-24
Improve generate-images & mass-update script and readme

Noki:


generate-images.sh
```diff
@@ -1,32 +1,100 @@
...
+# shellcheck disable=SC2053
+
+# Prepare options list by searching through locations/
+unset locations hosts
+
+# Extract valid locations
+mapfile -t locations < <(yq '.location' locations/*.yml | sed 's/\"//g' | sort)
+
+# Extract valid hosts
+mapfile -t hosts < <(yq '.hosts[] | select(.role == "corerouter" or .role == "gateway" or .role == "ap") | .hostname' locations/*.yml | sed 's/\"//g' | sort)
```
pktpls, line 11: You can use `yq -r` instead of the sed call: "-q  output raw strings, not JSON texts"

PR #1151, opened 2025-01-05
roles: remove all luci/uhttp packagages on low_mem / low_flash

Noki:
This goal of this PR is to reduce the image size for model files with low_mem and low_flash by removing luci packages and unused packages. However there are some issues

- the solution is not elegant and does not distinguish between APs and core routers
- the owm script does not work after this on core routers
- core routers flashed with this have issues to bring up SSH in a reasonable time, almost as randomness is missing. The situation resolves after several minutes or sometimes hours.


Noki:
we need to remove even more to make it work for hdk-78a and hdk-24... suggestions welcome

PolynomialDivision:
> we need to remove even more to make it work for hdk-78a and hdk-24... suggestions welcome

replace the routers xD

spolack:
IMO both options were a workaround to keep at least APs still supported, but for core routers i have doubts it's worth the effort.

PolynomialDivision:
Remove olsr since we have bird now.

PR #1160, opened 2025-01-09
Update k9 front house

iuljan:
Add usw-flex switch to front house to connect further private AP DAPx1860.

### DIFF
--- a/locations/k9.yml
+++ b/locations/k9.yml
@@ -21,6 +21,40 @@ hosts:
     role: ap
     model: "aruba_ap-303"
     wireless_profile: freifunk_default
+  - hostname: k9-ap-iuljan
+    role: ap
+    model: "dlink_dap-x1860-a1"
+    wireless_profile: freifunk_default
+  - hostname: k9-switch-front-house
+    role: ap
+    model: "ubnt_usw-flex"
+    poemgr_ports:
+      - name: lan2
+        port: 3
+        disabled: 1
+      - name: lan3  # PoE @ Loge
+        port: 2
+        disabled: 0
+      - name: lan4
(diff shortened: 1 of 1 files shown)


locations/k9.yml
```diff
@@ -21,6 +21,31 @@ hosts:
...
+        port: 2
+        disabled: 0
+      - name: lan4
+        port: 1
+        disabled: 1
+      - name: lan5
+        port: 0
+        disabled: 1
+
+    poemgr_power_budget: 24 #???
```
Noki, line 45: This setting is only required if you use active to passive converters that negotiate more power than they actually draw and therefore devices get not powered properly. With PoE only on one port you should not need this at all.

locations/k9.yml
```diff
@@ -21,6 +21,31 @@ hosts:
...
+    # custom switch config script. Should be applied with a mechanism, that
+    # includes files into root files system later on.
+    # host__rclocal__to_merge: #is it automatically executed? check ilr.yml
```
Noki, line 48: This can also be removed.

locations/k9.yml
```diff
@@ -21,6 +21,31 @@ hosts:
     role: ap
     model: "aruba_ap-303"
     wireless_profile: freifunk_default
+  - hostname: k9-ap-iuljan
+    role: ap
+    model: "dlink_dap-x1860-a1"
+    wireless_profile: freifunk_default
+  - hostname: k9-switch-front-house
+    role: ap #??
```
Noki, line 29: yes, AP is correct.

nailend:
@Noki thx for the review! I revised it.

PR #1167, opened 2025-01-16
tunspace: allow static IP instead of DHCP

pktpls:
context: freifunk-berlin/falter-packages#468


pktpls:
Buildbot will have finished the updated package feeds in the morning

Noki (approved):
Looks good to me

PR #1173, opened 2025-01-25
tempelwg: workaround for switch link issue on D-Link COVR-X1860

PolynomialDivision:
There is a periodic link up/down issue. As a workaround, disable EEE (Energy Efficient Ethernet).
See https://github.com/openwrt/openwrt/issues/17351 for more details.


maurerle:
Hi :) did you also have a continuous log spam, without the link ever coming up? Or just regular connectivity losses?
I did only experience this on 1 of 6 devices. Is the LAN port affected as well? :)

PR #1178, opened 2025-02-05
[WIP] core/ap: dsa config: fix nasty roaming bug

spolack:
replaces https://github.com/freifunk-berlin/bbb-configs/pull/1174

Still needs bisecting if and which of the following lines can be dropped:

* 3:	option vlan_filtering '1'
* 4:	option ports '{{ dsa_ports|join(' ') }}'
* 11:    {{ portmapping.append(port|string + (":t" if tagged else ":u*")) }}
* 14-21 ...

### DIFF
--- a/roles/cfg_openwrt/templates/common/config/dsa.network.inc
+++ b/roles/cfg_openwrt/templates/common/config/dsa.network.inc
@@ -1,13 +1,23 @@
 config device
 	option type 'bridge'
 	option name 'switch0'
+	option vlan_filtering '1'
+	option ports '{{ dsa_ports|join(' ') }}'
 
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
What about merging this as it is and later removing the unneeded stuff. Currently, I always need to cherry-pick the commits to have a working roaming setup because main is broken as you know. I would prefer a working config with maybe to much config over a broken one on main.

spolack:
Fix doesnt work for me, seeing stale MAC Entries again, this time on the Switch interface...

PolynomialDivision:
Indeed this is not fixing the issue.

pktpls:
This was continued and merged in #1264

PR #1180, opened 2025-02-05
scripts: use SSH connection check instead of ping for host verification

PolynomialDivision:
To run the mass update script, you must be connected to the internal mesh network. If you're not connected, the script will report that routers or access points cannot be found. However, many users connect via proxy jumps through our gateways to manage devices via SSH. In its current state, the script relies on ping to verify connectivity, which fails when SSH access is possible but ping is not working.

This update changes the script to check for an active SSH connection instead of relying on ping. This ensures that the SSH daemon is up and reachable, providing a more accurate verification of host availability, rather than just testing for ICMP echo requests.


Noki:
Thank you for this one! This is great! The only downside is that without a password prompt we can't update the old nanosations with this anymore as they don't support ed25519 and require rsa due to limitations of the tiny target. But creating an additional RSA key for that is probably the way to go and the tiny targets were always a bit problematic to flash anyway.

PR #1202, opened 2025-03-17
xdorf: switch core model

noxilixon:


Noki:
I suggest switching to an outdoor-router e.g. Cudy AX3000 so it covers more area and that is capable of bridging the distance towards rauchhaus in a single device. The goal should be to remove the 100 MBit limitation towards rauchhaus and to better cover the Wagenplatz e.g. by also adding one or two additional routers for that location.

noxilixon:
I think this is a good idea. I will forward it to the local maintainers. Still merging this pr, because it reflects the current situation.

PR #1203, opened 2025-03-21
rhnk: rename p2p devices and mesh networks

Noki:
This PR renames the p2p devices and corresponding mesh networks to make it easier to understand.


Noki:
got two thumbs up in matrix. ;-)

PR #1206, opened 2025-03-21
models: delete ubnt nanostation m5 xw

### LINKED ISSUE #1149: ubnt Nanostation M5 (XW): No Network connectivity after flashing bbb-configs
After flashing the device with the built image, it doesnt get network connectivity.

FFHener:
The networkconfig would need to be fixed in order to work. I think we could close #1149 as well.

In my opinion it doesnt makes sense to have this file in here in a non working state. In addition I decided not putting to much work into it as we are going to use more mordern hardware eventually. As no locations are using this hardware with bbb-configs right now I think we can safely delete it


xa-y-ax:
Ist da nicht noch so ein altes Ding am Vaterhaus? Sollten wir dann ersetzen.

PR #1218, opened 2025-04-17
all: change ieee80211w to required

Noki:
`ieee80211w: 1` does not work with quite some devices, mainly windows based devices but changing it to `ieee80211w: 2` always worked. I therefore suggest to change this to `ieee80211w: 2` as new default as it solved compatibility issues for me. Let me know if you have any concerns.


pktpls:
Yes let's do it :+1: Windows requires 2 for OWE.

PolynomialDivision (approved):

PR #1221, opened 2025-04-20
gateway: update firewall port openings

pktpls:
- Remove prometheus-poc (RIP)
- Remove radbahn mt76 debugging (RIP)
- Allow SSH for a magda client


PolynomialDivision (approved):

pktpls:
I think i will need to update that IPv6 address, after cleaning up at the magda side

Noki:
Maybe just add a host-network for magda as I did at my location.

pktpls:
> Maybe just add a host-network for magda as I did at my location.

Yep good idea, then I dont need to update the gateways if that little client machine decides to roll a different IPv6 address

PR #1237, opened 2025-05-06
Mlk nk new router

FFHener:


Noki:
Hmm, looks like I did not properly calculate the 802.11s metrics back in the day. Also included that fix here. Should be 4*512 and 5*512.

PR #1246, opened 2025-05-17
t1: remove unused snmp-device

FFHener:


Noki:
it's also still in the managment assignments

PR #1247, opened 2025-05-17
added snmp profiles for more switches

Noki:


PolynomialDivision:
scharni has also a netgear switch. should this work for it?

Noki:
Should work for most switches, yes. There seems to be a standard and most vendors use the same OIDs. Just try and if it does not work I will have a look.

PolynomialDivision:
The switch does not seem to have snmp support. :/

robertfoss:
LGTM

PR #1255, opened 2025-05-31
[RFC] templates: hacky roam bug fix

PolynomialDivision:


pktpls:
Btw the tunspace setting `uplink_mode = bridge` is called bridge because the macvlan child's operational mode is called bridge, not because of the wrapping bridge. Currently, the alternative option `direct` still also uses a wrapping bridge around the uplink interface.

pktpls:
> Can you give me config examples?

https://github.com/pktpls/bbb-configs/blob/ffb141a0f08aa8c06ae4ae5329ed7cfaa4c00e22/locations/pktpls.yml

I then went changed the uplink interface in /etc/config/network to be switch0.50 instead of a bridge, and changed uplink_ifname in /etc/config/tunspace to switch0.50. Additionally I tested the same with uplink_mode=direct.

PR #1259, opened 2025-06-06
[RFC] treewide: add linter and refactor code

PolynomialDivision:


pktpls:
We already had ansible-lint in the past, it was no fun. What problem are you trying to cover?

PolynomialDivision:
> We already had ansible-lint in the past, it was no fun. What problem are you trying to cover?

I would like to use language ansible serer in my ide ;) Why did we revert it, was it so awful?

pktpls:
> was it so awful

It did the most frustrating style checks and no functional checks. Stuff like "comments must have two spaces in front of `#`" or "task names must start with uppercase letter". 

It might have learned functional check in the meantime, I dont know. But if it's purely style checks, I'd vote against making it a required check for merging a PR.

I'm not against enforcing code style, but it should be an autoformatting tool (like `shfmt` or `go fmt`) that can be run manually at will. I really really hate when a PR gets rejected because of whitespace :-)

falter-packages for example has `make lint` which is a PR check, and `make fmt` which is a manual thing.

roles/cfg_openwrt/tasks/imagebuilder.yml
```diff
@@ -105,6 +105,7 @@
     path: "{{ configs_dir }}/etc/apk/repositories.d/falter.list"
     line: "{{ feed }}"
     create: true
+    mode: "0644"
```
pktpls, line 108: `mode` seems to be an actual functional check. This I find acceptable to result in a check fail blocking a PR from being merged.

But all the other stuff is just formatting and shouldn't be part of a "lint" task, I think.

PolynomialDivision:
> I'm not against enforcing code style, but it should be an autoformatting tool (like `shfmt` or `go fmt`) that can be run manually at will. I really really hate when a PR gets rejected because of whitespace :-)

But you can just run `ansible-lint --fix` and it fixes everything for you. And even better you can add it to your IDE for automatic formating. I love such tools because it bothers me doing all fixes by hand.

PolynomialDivision:
So I'm fine with removing: https://github.com/freifunk-berlin/bbb-configs/pull/1259/commits/600b101099d428c3f0d6b86a0ccbd0025899595f

pktpls:
> But you can just run ansible-lint --fix and it fixes everything for you

Ah that's new, very good!

We really need a makefile that does exactly what the github checks do, just locally

roles/cfg_openwrt/templates/common/config/dsa.network.inc
```diff
@@ -6,11 +6,11 @@ config device
   {% set portmapping = [] %}
   {% for port in dsa_ports %}
     {% set tagged = not network.get('untagged') %}
-    {{ portmapping.append(port|string + (":t" if tagged else "")) }}
-  {%- endfor %}
+    {% set _ = portmapping.append(port ~ (":t" if tagged else "")) %}
+  {% endfor %}
 
-config bridge-vlan 'vlan_{{ network['vid'] }}'
```
PolynomialDivision, line 12: here we dont name the bridge vlans anymore (seems unnecessary)

PolynomialDivision:
No worries, I didnt expect this to be merged directly. I need to move some fixes from the last commit into a separate ansible-lint fix, as ansible-lint wasn’t able to fix them automatically. Let me review everything.

Actually, I just wanted to clean up the code before revisiting the roaming config. Also, there is a significant difference in the network configuration: we dont add a separate device bridge. I believe this might not be necessary, but unfortunately, for the "int_port" devices, it seems to be required. However, it should be removed in this or in future commit.

I will write [RFC] to commit message so it gets clear I just want feedback for now.

PR #1261, opened 2025-06-12
treewide: enable mesh_nolearn

PolynomialDivision:
Disable 802.11s path discovery because we run our own mesh routing daemon.


PolynomialDivision:
So this is an approve (?) so I will merge. :)

PR #1264, opened 2025-06-23
templates: fix roaming bug by avoiding bridge in DSA setups

PolynomialDivision:
Only use bridges with switch ports devices. Remove bridge in DSA configurations, as it causes roaming issues.

Maybe before checking int_port we could also check for dsa_ports.

### DIFF
--- a/roles/cfg_openwrt/templates/common/config/dsa.network.inc
+++ b/roles/cfg_openwrt/templates/common/config/dsa.network.inc
@@ -6,7 +6,7 @@ config device
   {% set portmapping = [] %}
   {% for port in dsa_ports %}
     {% set tagged = not network.get('untagged') %}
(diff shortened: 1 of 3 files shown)


PolynomialDivision:
@spolack Any objections?

pktpls:
Cool :+1: Thanks for your patience :)

I couldnt think of a good place to test the fix, so I mainly just checked it doesnt break other setups.

- hway: x86 corerouter, single AP with DSA, `dhcp` tagged, `prdhcp` tagged but without wifi
- lause: DSA corerouter, single AP with DSA, `dhcp` tagged, tunspace `uplink` untagged
- muggel: DSA corerouter, no other devices, tunspace `uplink` via LTE USB stick

PR #1270, opened 2025-06-27
c-base: replaced nanobeam with af60lr

nicolasberens:
let's prefer the af60 and keep the waveap as a backup to see if the performance is sufficient


nicolasberens:
should be fixed now

PR #1277, opened 2025-07-02
khsb2:init

riskieee:


noxilixon:
Hi riskieee :) Nice that you created  this location. You used the same IPv6 address `2001:bf7:840:e00::/56` as the location [glanz15](https://github.com/freifunk-berlin/bbb-configs/blob/main/locations/glanz15.yml). This needs to be fixed. Also I was told that you will get new IPv4 addresses. I changed this PR to a draft for now.

PR #1278, opened 2025-07-04
templates: add option to enable flow offloading

nicolasberens:
Add option to enable flow offloading:

https://openwrt.org/docs/guide-user/perf_and_log/flow_offloading

this massively improves the routing performance on several devices

e.g.:
- Fritzbox 7530 from ~250 Mbits bidirectional to 500 (limited by the cpu that is connected only via gigabit to the switch)
- EX400 from ~200Mbits to ~950 bidirectional when using both interfaces

Because of the pitfalls mentioned in the wiki page i would like to test this on a few selected sites

### DIFF
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -59,4 +59,3 @@ mesh_metric_adhoc_11g_standard: 2560
 # https://openwrt.org/docs/guide-user/perf_and_log/flow_offloading
 # Possible values: none, sw, hw
 flow_offload: none
-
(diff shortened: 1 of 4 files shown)


pktpls:
Total off-topic, aber:
image
Ich finds nicht gut dass jetzt wieder Unwichtigkeiten wie Newlines zu roten Builds führen :( In meinen eigenen PRs ist sowas der absolute Motivationskiller. Das ist genau warum ich den Linter damals entfernt habe.

pktpls:
On-topic: I approve :)

PR #1296, opened 2025-07-19
model: do single-port DSA properly on Zyxel

pktpls:
NWA50AX tested with lause-ap1 and hway-indoor.
NWA55AXE tested with huette-core.

There's a few other devices that should be changed to `dsa_ports`, e.g. the Cudys and D-Links. Others already do it correctly, e.g. Unifi 6 Lite.


PolynomialDivision:
maybe it has something to do with my commit. probably we should check for single port dsa devices.

PolynomialDivision:
`brctl` is legacy, try  looking at the output of `bridge vlan show`

PR #1303, opened 2025-07-24
model: add tp-link cpe710 v2

ZuquetuPlexis:


pktpls:
Looks good, but needs `openwrt_version: 24.10-SNAPSHOT` i think

ZuquetuPlexis:
> Looks good, but needs `openwrt_version: 24.10-SNAPSHOT` i think

Why does it need the snapshot version?

PR #1312, opened 2025-08-01
imageprofile: set 24.10-SNAPSHOT as new default

pktpls:
The end-of-life of 23.05 is scheduled for Monday, it's about time we upgrade the remaining devices to 24.10.

This PR flips the default from 23.05 to 24.10, which makes it easier to grep for 23.05 devices.


Noki (approved):
looks good to me

PR #1322, opened 2025-08-08
templates: generate unique deterministic MACs for mesh interfaces

Noki:


PolynomialDivision:
Please put that into the commit message, so we can still understand the reasoning behind this change later on.

PR #1333, opened 2025-08-22
templates: babel metrics per neighbor

Noki:
This PR introduces babel metrics per neighbor. The idea is to be able to add costs for individual neighbors based on their link local IP. This is useful for 802.11s mesh interfaces or APs with multiple stations. To see the neighbor IPs use:

birdc show babel neighbors


pktpls:
It's puzzling that it doesn't work with `gw`. On the CLI it matches, e.g. `birdc show route where 'gw = fe80::58:d6ff:fecb:dad0'`.

PR #1339, opened 2025-08-28
model: use 24.10 instead of snapshot for cudy models

Noki:
The Cudy models are supported in 24.10 for some time now. This PR switches the models from snapshot to 24.10 for more stable builds. Users can keep using snapshot for individual locations by specifying it in the location yml file.


pktpls:
Yep makes sense! :+1:

PR #1347, opened 2025-09-03
strom: switched eth-ports

FFHener:
This is needed as they got switched in the 24.X release

- [x] Deploy https://github.com/freifunk-berlin/meta/issues/168


pktpls:
And best to add a comment about the connection being over the Intel NIC, instead of the Mellanox NIC like the other locations

PR #1360, opened 2025-09-17
owrt one: set static MAC

FFHener:
Otherwise a random one gets generated

### DIFF
--- a/group_vars/model_openwrt_one.yml
+++ b/group_vars/model_openwrt_one.yml
@@ -7,6 +7,8 @@ dsa_ports:
   - eth0
   - eth1
 
+requires_mac_override: true
+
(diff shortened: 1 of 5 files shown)


Noki:
> The problem is with the tunspace package. It generates a random mac address [here](https://github.com/freifunk-berlin/falter-packages/blob/main/packages/tunspace/tunspace.uc#L459). This needs to be changed to fix the underlying problem

 @pktpls can we have an option for tunspace to generate a deterministic MAC address based on the hostname + interface, as I have done in this commit (https://github.com/freifunk-berlin/bbb-configs/commit/083356580e0644e262236c3db3a12d484e0d5903)?

PR #1361, opened 2025-09-17
Fix ansible

FFHener:


requirements.txt
```diff
@@ -1,10 +1,9 @@
-ansible >= 2.13
-ansible-lint >= 25.6.1
-ansible_merge_vars >= 5.0.0
-jmespath >= 1.0.1
-netaddr >= 0.8.0
-yamllint >= 1.30.0
-ansible-lint >= 6.22.1
```
Noki, line 7: ansible-lint is a requirement for the generate-images.sh script, which supports linting with ansible lint.

PR #1366, opened 2025-09-22
hacrafu-eliche: init location, replace falter

ZuquetuPlexis:
This PR should be merged once mesh-link is reliably prefered over wan uplink

- [x] check if wan uplink is used as backup only

### DIFF
--- a/locations/hacrafu-eliche.yml
+++ b/locations/hacrafu-eliche.yml
@@ -1,0 +1,110 @@
+---
+
+location: hacrafu-eliche
+location_nice: Angerscheune e.V., Dorfpl. 1, 15370 Petershagen
+latitude: 52.52354918402503
+longitude: 13.769907550925465
+contact_name: "Hacken Craften Funken e.V."
+contact_nickname: "HaCraFu e.V."
+contacts:
+  - "freifunk@hacrafu.de"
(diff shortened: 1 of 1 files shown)


locations/hacrafu-eliche.yml
```diff
@@ -0,0 +1,110 @@
...
+# TUNNEL    10.248.84.160/31
+
+networks:
+  # MESH - 5 GHz 802.11s
+  - vid: 20
+    role: mesh
+    name: mesh_5g
+    prefix: 10.248.76.126/32
+    ipv6_subprefix: -20
+    mesh_metric: 128
```
FFHener: Ist there a reason to set this metric here this good?

ZuquetuPlexis: > Ist there a reason to set this metric here? The [default babel metrics](https://github.com/freifunk-berlin/bbb-configs/blob/main/group_vars/all/general.yml#L48-L56) allready prioritize Meshlinks over VPN links. 


yes, there is no LOS to this device. I tested a lot of different mesh metric settings, but with the default metric the wan uplink got prefered every time.
This configuration seems to give the best results but i'm  still testing the (mesh) connection.

PR #1378, opened 2025-10-07
models: remove banana bpi r2

PolynomialDivision:
There is no location using the bpi r2.


PolynomialDivision:
I think noone will ever use this device. So it is one target less we have to bump to 24.10.

pktpls:
image


My favorite code is deleted code :+1: 

We can bring it back in the future if someone needs it

pktpls:
@PolynomialDivision also remove `group_vars/target_mediatek_mt7623`, this was the only model using that

PR #1393, opened 2025-10-17
gruni73: add bar ap and upgrade to 24.10

PolynomialDivision:
Add the bar AP and update belkins to 24.10.

### DIFF
--- a/locations/gruni73.yml
+++ b/locations/gruni73.yml
@@ -27,6 +27,12 @@ hosts:
     role: ap
     model: "ubnt_rocket-5ac-lite"
 
+  - hostname: gruni73-zielona
+    role: ap
+    model: "linksys_e8450-ubi"
+    openwrt_version: 24.10-SNAPSHOT
+    wireless_profile: gruni73
+
 snmp_devices:
   - hostname: gruni73-switch
     address: 10.31.156.2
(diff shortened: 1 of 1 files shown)


locations/gruni73.yml
```diff
@@ -123,8 +130,58 @@ networks:
     mesh_iface: mesh
     mesh_metric_lqm:
       - default 0.4 # penalty on wild meshing
+  - vid: 50
+    name: prdhcp
+    role: uplink
```
pktpls: Need `role: dhcp` here and the convention is vlan 41 - uplink and 50 would be for tunspace

PolynomialDivision: dhcp does not work as rule, since it expects a prefix and so on. I just want to have a untagged switch bridge.

pktpls: Ah - then use `role: ext` with untagged :)

PolynomialDivision: ah thanks! and vlan?

PolynomialDivision: thanks. i would prefer untagged: true since i just wan all ports to be untagged, regardeless what router is used and to what router we will siwtch in the future.
I think the comment is wrong in this examples. ;) no isolation und so is happening there.

Noki (approved):

PR #1405, opened 2025-10-23
khsb2: fix ip range overlap

Noki:
I think this is what was planned...


pktpls:
@khsbsolc @FFHener you need to flash khsb2-core wit the fixed DHCP prefix. Merge when it's done

PR #1411, opened 2025-10-26
gateway: allow incoming traffic to mirror.freifunk.dev

pktpls:


pktpls:
This has already been applied to the gateways manually (edit /etc/config/firewall and run reload_config)

PR #1415, opened 2025-10-27
mahalle: mesh walde

noxilixon:


FFHener (changes requested):
You only need airos-dfs-reset on the side of the AP. The AP defines the frequency. The other side will scan for the correct frequency

PR #1416, opened 2025-10-27
magda: add open network to remise without client isolation

FFHener:
This is needed to connect devices locally with eachother

### DIFF
--- a/locations/magda.yml
+++ b/locations/magda.yml
@@ -16,6 +16,7 @@ hosts:
   - hostname: magda-ap-remise
     role: ap
     model: "zyxel_nwa50ax"
+    wireless_profile: remise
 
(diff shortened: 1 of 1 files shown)


locations/magda.yml
```diff
@@ -102,35 +122,52 @@ networks:
...
+        encryption: owe
+        network: dhcp
+        radio: [11a_standard, 11g_standard]
+        ifname_hint: ffowe
+        ieee80211w: 2
+
+      - mode: ap
+        ssid: Remise
+        encryption: none
+        network: prdhcp
```
pktpls: Wrong `network` and `ifname_hint`  (should be `remise` I guess)

LGTM otherwise :+1:

PR #1424, opened 2025-10-29
model: remove all unused device models

pktpls:
```sh
> ls group_vars/model_* | xargs -n1 basename -s .yml | sed -e 's/^model_//' | sed -e 's/_/./g' | while read m ; do grep -P "$m" locations/*.yml >/dev/null || ls group_vars/model_* | grep -P "$m" ; done | xargs git rm
```


ZuquetuPlexis:
I have a fork where i use 7 of the listed models in different branches. Should i just recreate the modelfiles for the branches of my fork?

PR #1427, opened 2025-11-01
Ansible Lint Workaround

ZuquetuPlexis:


Noki:
Did it with a slightly different approach in #1428

PR #1437, opened 2025-11-05
templates:dhcp: add hostname to config host section

### LINKED ISSUE #1433: dhcp config file: add name to host entries
currently a host in the dhcp config looks like this:

```
config host
	option name 'myhostname'
	option ip '10.20.30.40'
```

It would be very helpful, especially in custom rc.local configs to have the following

```
config host 'myhostname'
	option name 'myhostname'
	option ip '10.20.30.40'
```

The specific use case that I have is that I need to set up a static dhcp lease.  Instead of looping through all possible host entries until 'myhostname' is found or guessing it will work with ```dhcp.@host[5].mac``` since the index number could change.  It would be al lot simpler to just do ```uci set dhcp.myhostname.mac=aa:bb:cc:dd:ee:ff```

FFHener:
This fixes #1433

I didnt added it to the gateways, only core routers, as they dont have `config host` in them. 

Before:
```
config host
	option name 'rhnk-core'
	option ip '10.230.3.65'
```

After:
```
config host 'rhnk-core'
	option name 'rhnk-core'
	option ip '10.230.3.65'
```

I couldnt find this syntax in the [official documentation](https://openwrt.org/docs/guide-user/base-system/dhcp?s[]=0&s[]=0.#static_leases), but assume it should be right as it was working before with @pmelange

### DIFF
--- a/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
+++ b/roles/cfg_openwrt/templates/corerouter/config/dhcp.j2
@@ -64,7 +64,7 @@ config dhcp 'dhcp_{{ name }}'
 
 {% for network in networks | selectattr('role','equalto','dhcp') %}
     {% for assignment in network['assignments'] %}
-config host
+config host '{{ assignment.replace("-", "_") }}'
 	option name '{{ assignment }}'
 	option ip '{{ network['prefix'] | ansible.utils.ipaddr(network['assignments'][assignment]) | ansible.utils.ipaddr('address') }}'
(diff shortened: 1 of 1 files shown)


pmelange:
A section can be named or not, and it doesn't matter for dhcp.  We could also name the sections in other config files, such as olsrd, to make it easier for scripting there too.  The init script will just loop through all the "host" config sections and doesn't care about the name.

pmelange (commented):
small change suggestion

PR #1438, opened 2025-11-05
template:tunspace: add deterministic mac

FFHener:
Add option in configs for a deterministic mac similar to [existing approaches](https://github.com/freifunk-berlin/bbb-configs/blob/512bdbe29fb7b76256b3dac50ad70a592772c614/roles/cfg_openwrt/templates/common/config/network.j2#L57-L60). This option can later be used in tunspace package. [Corresponding change there](https://github.com/freifunk-berlin/falter-packages/pull/493)


PolynomialDivision:
Can you maybe shortly describe in the commit why this is useful?

PR #1443, opened 2025-11-06
add scherer8-abb

pmelange:
Is flashed and running

### DIFF
--- a/locations/scherer8-abb.yml
+++ b/locations/scherer8-abb.yml
@@ -1,0 +1,167 @@
+---
+location: scherer8-abb
+location_nice: Schererstr 8
+latitude: 52.548006693797761
+longitude: 13.366858363151452
+
+contact_nickname: "Perry"
+contacts:
+  - "isprotejesvalkata [attt] gmail com"
+
+# This is a mesh-only node, but we want to keep it on the map so it's
+# configued as a core router.  Huh
+hosts:
+  - hostname: scherer8-abb
+    role: corerouter
+    model: "tplink_tl-wdr4900-v1"
+    wireless_profile: scherer8-abb
+    host__rclocal__to_merge:
+      - |
+        # set up the ports correctly
+        # vlan 211
+        uci set network.vlan_211.ports="wan:t lan1:t"
+        # vlan 10
+        uci set network.vlan_10.ports="wan:t lan1:u"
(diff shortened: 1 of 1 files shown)


locations/scherer8-abb.yml
```diff
@@ -0,0 +1,167 @@
...
+    mesh_radio: 11g_standard
+    mesh_iface: mesh
+
+  # freifunk client network
+  - vid: 10
+    role: ext
+    name: freifunk
+    inbound_filtering: false
+    enforce_client_isolation: false
+    no_corerouter_dns_record: true
```
PolynomialDivision: role ext seems wrong for the freifunk network?

pmelange: The dhcp network for freifunk is controlled by the real core router of the network (scherer8.ff).  The point here is to have an unmanaged vlan which gives wifi access to the dhcp network provided by the core router.  Therefore ext.

locations/scherer8-abb.yml
```diff
@@ -0,0 +1,167 @@
...
+          uci set firewall.zone_hauslan=zone
+          uci set firewall.zone_hauslan.name=hauslan
+          uci set firewall.zone_hauslan.forward=ACCEPT
+          uci add_list firewall.zone_hauslan.network=hauslan
+          uci set firewall.zone_transtor=zone
+          uci set firewall.zone_transtor.name=transtor
+          uci set firewall.zone_transtor.forward=ACCEPT
+          uci add_list firewall.zone_transtor.network=transtor
+          uci commit firewall
+        fi
```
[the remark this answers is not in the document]
pmelange: The autoconfig does not put the "ext" networks into any kind of zone.  Therefore I create the needed zones for the created ext networks.  Specifically, the freifunk wifi "ext" needs to be added to the freifunk zone.

locations/scherer8-abb.yml
PolynomialDivision: Than you should not add multiple locations, instead you just add the APs into the scherer8 location, as role: ap? Than the dhcp role should work as expected.

PolynomialDivision:
Okay, I am not completely happy with the uci workarounds, but it is okay for me. So you can just see my remarks as more an advice than forcing you to change anything.

PR #1444, opened 2025-11-07
perleberger36: add 60ghz waveap

nicolasberens:
@pmelange

### DIFF
--- a/locations/perleberger36.yml
+++ b/locations/perleberger36.yml
@@ -20,10 +20,10 @@ hosts:
         uci set network.vlan_11.ports='wan:u lan2:u lan3:u lan5:u'
         uci commit network
         reload_config
-    
+
 
 snmp_devices:
-  - hostname: perleberger36-strom2a
+  - hostname: perleberger36-ssw
     address: 10.36.243.162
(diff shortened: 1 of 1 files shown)


locations/perleberger36.yml
```diff
@@ -13,6 +13,13 @@ hosts:
     role: corerouter
     model: "mikrotik_routerboard-750gr3"
     wireless_profile: freifunk_default
+    host__rclocal__to_merge:
+      - |
+        uci set network.vlan_105.ports='wan:t lan2:t lan3:t lan4:u lan5:t'
```
pmelange: Let's add a comment like
```
# The af60 is connected directly to lan4 which needs to be untagged and also removed from vlan_11
```

locations/perleberger36.yml
```diff
@@ -13,6 +13,13 @@ hosts:
...
+        uci set network.vlan_11.ports='wan:u lan2:u lan3:u lan5:u'
+        uci commit network
+        reload_config
+
 
 snmp_devices:
   - hostname: perleberger36-strom2a
```
pmelange: please rename to ssw

PR #1451, opened 2025-11-12
Rauchhaus and Xdorf current state

noxilixon:
The model was changed at the location without a change in the config. I now changed this to represent the current state of rauchhaus and xdorf.


locations/rauchhaus.yml
```diff
@@ -118,6 +122,7 @@ networks:
       rauchhaus-basement: 15
 
 location__channel_assignments_11a_standard__to_merge:
+  rauchhaus-core: 36-20
```
FFHener: Is there a reason to reuse the channel so much instead of spreading it out?

PR #1456, opened 2025-11-15
segen: refresh

nicolasberens:
https://github.com/freifunk-berlin/meta/issues/187

- Replace unused weißensee & no antennas with a sector (either rocket or isostation) to connect kollage kollektiv
- Move BHT connection to fardf, since BHT is going away

### DIFF
--- a/locations/segen.yml
+++ b/locations/segen.yml
@@ -52,7 +52,7 @@ snmp_devices:
     address: 10.31.6.13
     snmp_profile: airos_6
 
-  - hostname: segen-bht
+  - hostname: segen-fardf
     address: 10.31.6.14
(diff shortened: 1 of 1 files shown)


locations/segen.yml
```diff
@@ -128,7 +129,9 @@ networks:
     ipv6_subprefix: 1
     assignments:
       segen-core: 1
-      segen-switch: 2
+      segen-switch-1: 2
+      segen-switch-2: 3
+      segen-switch-octagon: 3
```
pmelange: These both have the same assignment.  If that works as expected (I have no idea), then maybe it's good to add a comment so it's understood in the future.

PR #1483, opened 2025-12-17
ohlauer: new IPs from Inter.Link

pktpls:
Ohlauer is already back online. Only the whois database entry for IPv4 is missing, we want that so we can handle abuse emails ourselves.

Also reverted all the Protectli device commits, the device change needs to wait until January.

The new uplink settings are:
- 2a11:4141:a402::5/127
- 45.153.82.209/31
- AS5405

### DIFF
--- a/group_vars/role_gateway/general.yml
+++ b/group_vars/role_gateway/general.yml
@@ -18,9 +18,6 @@ freifunk_wahlkreis_prefixes:
 freifunk_wahlkreis_announcement_prefix: 44
 # Change to 48 if you wnat to announce 48s
 
-local_asn: 44194
-peer_asn: 25291
-
 role_gateway__sysctl__to_merge:
   net.netfilter.nf_conntrack_max: 524288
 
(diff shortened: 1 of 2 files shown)


locations/ohlauer.yml
```diff
@@ -1,21 +1,21 @@
...
 
 location: ohlauer
-location_nice: Ohlauer Str. 43, 10999 Berlin
+location_nice: Ohlauer
 latitude: 52.49378703855491
 longitude: 13.42969238683751
 altitude: 55
 community: true
 
+local_asn: 44194
```
FFHener: This var should be set for all gaateways

FFHener: I see your point, but whats about the other gateways? If we remove the variable from the group vars we need to add it to every gateway? Or is this the only one where we currently announce this ASN?

pktpls: Ah that's what you mean, yes the other three (ak36, l105, strom) already have their own `local_asn`+`peer_asn`, so with this PR it's consistent

PR #1488, opened 2025-12-24
model: bump some targets to 25.12

PolynomialDivision:


FFHener:
Have you tested all of them?

PR #1498, opened 2026-01-04
model: zyxel-nbg6617: update to OpenWrt 25.12

PolynomialDivision:
Update to newest OpenWrt Release.


techn0punk:
The new version is running for some hours now on emser97 without problems.

PR #1505, opened 2026-01-09
Wikiupdater: refactor

FFHener:
As the new community flag was introduced i took the time and refactored the wikiupdater. Its using the newly introduced community flag variables. In addition I got rid of about half of the lines of code and the output is prettier. I didnt changed any functionality.


pktpls:
LGTM apart from that one thing :+1:

PR #1511, opened 2026-01-17
model: use our mirror for 23.05-SNAPSHOT

pktpls:
23.05-SNAPSHOT has been deleted from downloads.openwrt.org, it's been EOL for a few months.


pktpls:
Also added the ability to configure the OpenWrt mirror URL

Noki:
At least for jup-bullet-ap4 this does not work: `Profile "ubnt_bullet-m-ar7241" does not exist!`

pktpls:
I switch 23.05 to use our mirror.

We can fix the images being too big separately. It must be from something else we did in the past couple of months...

PR #1514, opened 2026-01-21
ansible: switch to merge_variables lookup filter

pktpls:
This new filter is part of the ansible.community.general collection since the collection's version 6.5.0.

Two advantages:

1. No more need for an action plugin.

2. We get more control over the key names, while the old action plugin only allowed for suffix to be configured. We could now do a full regexp and achieve shorter key names such as `packages__bird` instead of the unwieldy `bird__packages__to_merge`.

I verified that the configs for all hosts build the same as before.

### DIFF
--- a/action_plugins/merge_vars.py
+++ b/action_plugins/merge_vars.py
@@ -1,1 +1,0 @@
-from ansible_merge_vars import ActionModule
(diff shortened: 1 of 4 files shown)


pktpls:
:clown_face: :clown_face: :clown_face: How helpful... the linter gained a passive-aggressive tone...
```
Error: Jinja2 spacing could be improved: {{ lookup('community.general.merge_variables', '__'+item.key+'__to_merge', pattern_type='suffix', initial_value=item.value) }} -> {{ lookup('community.general.merge_variables', '__' + item.key + '__to_merge', pattern_type='suffix', initial_value=item.value) }}
```

PR #1515, opened 2026-01-21
More debugging tools

pktpls:
- Now also prints the make command if you only generate the config, but don't build the image
- Writes a JSON dump of `hostvars` to `/tmp/images/<hostname>.json`.

### DIFF
--- a/roles/cfg_openwrt/tasks/imagebuilder.yml
+++ b/roles/cfg_openwrt/tasks/imagebuilder.yml
@@ -1,10 +1,4 @@
 ---
-- name: Create images dir
-  file:
-    state: directory
-    path: "{{ images_dir }}"
-    mode: "755"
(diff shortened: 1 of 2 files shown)


FFHener (dismissed):
[The definition of the variable](https://github.com/freifunk-berlin/bbb-configs/blob/main/roles/cfg_openwrt/tasks/imagebuilder.yml#L64-L75) should be removed as well

pktpls:
> [The definition of the variable](https://github.com/freifunk-berlin/bbb-configs/blob/main/roles/cfg_openwrt/tasks/imagebuilder.yml#L64-L75) should be removed as well

I'm an idiot, we do of course still need that variable - what are we going to replace \_\_INSTR_SET\_\_ with? :)

PR #1516, opened 2026-01-21
imageprofile: invert the conditional_packages logic

pktpls:
The goal of this PR is to have all packages, even conditional packages, defined in the imageprofile files. This will prevent future confusion as to why certain packages are included or not included, and where that is defined.

To achieve this, we put all groups of conditional packages into respective new merge_vars variables in the imageprofile files. Then later in conditional_vars.yml we remove these based on the given conditions.

- [ ] Double-check the inversion of the individual conditionals
- [ ] Add more comments

### DIFF
--- a/group_vars/all/imageprofile.yml
+++ b/group_vars/all/imageprofile.yml
@@ -12,18 +12,19 @@ imagebuilder_filename: "openwrt-imagebuilder-{{ openwrt_version }}-{{ target | r
 imagebuilder: "{{ openwrt_mirror }}/{{ openwrt_base }}/targets/{{ target }}/{{ imagebuilder_filename }}"
 feed: "src/gz openwrt_falter https://firmware.berlin.freifunk.net/feed/{{ feed_version }}/packages/__INSTR_SET__/falter"
 
+# The packages list is spread over a number of lists for two reasons:
+# 1. Individual locations want to override particular packages
+#    without dealing with the complete packages list.
+# 2. We want to omit certain packages based on device properties or config,
+#    e.g. devices with low ram / low flash, or with or without tunspace uplink.
+
 all__packages__to_merge:
-  - ethtool
   - ip6tables-nft # Its not pulled in by default anymore bc fw4
   - iptables-nft
-  - iperf3
-  - iwinfo
   - ip
   - kmod-nft-bridge
   - mtr
   - nftables
-  - tcpdump-mini
-  - vnstat
   - -ppp
   - -ppp-mod-pppoe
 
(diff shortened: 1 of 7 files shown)


group_vars/all/imageprofile.yml
```diff
@@ -12,18 +12,19 @@ imagebuilder_filename: "openwrt-imagebuilder-{{ openwrt_version }}-{{ target | r
...
+
 all__packages__to_merge:
-  - ethtool
   - ip6tables-nft # Its not pulled in by default anymore bc fw4
   - iptables-nft
-  - iperf3
-  - iwinfo
   - ip
   - kmod-nft-bridge
   - mtr
```
Noki, line 26: mtr should also go into `notlow__packages__to_merge`

roles/cfg_openwrt/tasks/conditional_packages.yml
```diff
@@ -1,103 +1,70 @@
...
+    low__packages__to_merge: []
+  when: not (low_mem | default(false) or low_flash | default(false))
 
-- name: "Disable uhttpd on low mem"
+- name: "Omit LuCI if low mem or low flash unless corerouter"
   set_fact:
-    disabled_services: "{{ disabled_services + ['uhttpd'] }}"
-  when: low_mem | default(false)
+    all_luci_base__packages__to_merge: []
+  when: (low_mem | default(false) or low_flash | default(false)) and role != "corerouter"
```
Noki, line 53: The LuCI stuff is huge and for low_mem and low_flash routers we should also remove it from the corerouter. uhttpd will already be disabled on low_mem so it will not be used. We should either have a setting like `no_web` that allows us to set this manually or just omit for corerouters. I think the only reason for not omitting at the moment is the OWM script, but we can simply use the ucode rewrite (https://github.com/freifunk-berlin/falter-packages/pull/519) there once merged.

group_vars/all/imageprofile.yml
```diff
@@ -12,18 +12,19 @@ imagebuilder_filename: "openwrt-imagebuilder-{{ openwrt_version }}-{{ target | r
 imagebuilder: "{{ openwrt_mirror }}/{{ openwrt_base }}/targets/{{ target }}/{{ imagebuilder_filename }}"
 feed: "src/gz openwrt_falter https://firmware.berlin.freifunk.net/feed/{{ feed_version }}/packages/__INSTR_SET__/falter"
 
+# The packages list is spread over a number of lists for two reasons:
+# 1. Individual locations want to override particular packages
+#    without dealing with the complete packages list.
+# 2. We want to omit certain packages based on device properties or config,
+#    e.g. devices with low ram / low flash, or with or without tunspace uplink.
+
```
Noki, line 20: Please add a reference to conditional_packages.yml for controlling how we build package lists and disabled services lists so it becomes clear where the actual magic happens.

group_vars/all/imageprofile.yml
```diff
@@ -46,5 +47,47 @@ all_collectd__packages__to_merge:
   - collectd-mod-ping
   - collectd-mod-uptime
 
-all__disabled_services__to_merge:
-  - "olsrd6"
+multicore__packages__to_merge:
+  - irqbalance
+
+# Will be removed unless wifi_roaming
```
Noki, line 53: Remove the comment. A reference to conditional_packages.yml at the top should be enough to understand the logic.

group_vars/all/imageprofile.yml
```diff
@@ -46,5 +47,47 @@ all_collectd__packages__to_merge:
...
+dawn__packages__to_merge:
+  - dawn
+  - luci-app-dawn
+  - umdns
+
+# Will be removed unless it's a device with DSA switch driver
```
Noki, line 59: Remove the comment. A reference to conditional_packages.yml at the top should be enough to understand the logic.

group_vars/all/imageprofile.yml
```diff
@@ -46,5 +47,47 @@ all_collectd__packages__to_merge:
...
+  - -kmod-phy-ath79-usb
+  - -kmod-usb-core
+  - -kmod-usb-ehci
+  - -kmod-usb-ohci
+  - -kmod-usb2
+
+qos__packages__to_merge:
+  - qos-scripts
+
+# This will be overridden unless low_mem
```
Noki, line 91: Remove the comment. A reference to conditional_packages.yml at the top should be enough to understand the logic.

Noki (changes requested):
Looks quite good. Hope we can get this merged soon. Will be quite an improvement!

Noki:
I think roles/cfg_openwrt/tasks/merge_vars.yml also needs an overhaul. I would expect that the conditional packages get merged first and afterwards host or model specific settings are merged. I think order matters here.

pktpls:
> I think roles/cfg_openwrt/tasks/merge_vars.yml also needs an overhaul. I would expect that the conditional packages get merged first and afterwards host or model specific settings are merged. I think order matters here.

It already got an overhaul, what's on your mind?

pktpls:
> It already got an overhaul, what's on your mind?

Ah yes I understand, but I'm not sure I agree merge ordering is important. Have we ever had a problem related to that?

PR #1538, opened 2026-02-18
collectd: add ping monitoring for all DNS servers

Noki:


FFHener:
I added this change to `lause-core`. I am wondering if there is a benefit to add all alternative addresses of the same DNS server or if this just clutters the monitoring

PR #1584, opened 2026-04-06
Update dorfplatz.yml (added/ changed hosts)

unholyscylla:
i changed the device model of one host
i added two more hosts so we can put more APs

### DIFF
--- a/locations/dorfplatz.yml
+++ b/locations/dorfplatz.yml
@@ -5,7 +5,7 @@ location_nice: Dorfplatz
 latitude: 52.517924
 longitude: 13.457358
 altitude: 37
-community: berlin
+community: true
 
 hosts:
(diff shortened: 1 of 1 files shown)


locations/dorfplatz.yml
```diff
@@ -5,7 +5,7 @@ location_nice: Dorfplatz
 latitude: 52.517924
 longitude: 13.457358
 altitude: 37
-community: berlin
+community: true
```
FFHener, line 8: This flag got changed, so it should stay `berlin`

FFHener (changes requested):

locations/dorfplatz.yml
Noki: @FFHener we should create a CI check for that... ;-)

PR #1591, opened 2026-04-12
feat: wireguard-access via corerouters

Noki:
Add per-VLAN WireGuard access VPN to corerouters. Each peer is assigned to a specific VLAN/network via the 'vid' field, with IP addresses derived from the network's assignments. This provides access and behaves exactly as if connected via WiFi/LAN to that VLAN.

Key features:
- Per-VLAN WireGuard interfaces (e.g., wg_wexample-laptop-1)
- Zone placement based on network's inbound_filtering setting
- Gateway firewall rules allow incoming WG connections from mesh
- Validation errors for missing required fields
- Documentation with examples

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -295,6 +295,105 @@ An optional configuration parameter can be added to any non-tunnel interface to 
 
 **WARNING:** This is made ineffective if [flow offloading](#flow-offloading) is enabled.
 
+### wireguard access vpn
+
+The core router can act as a WireGuard server, allowing remote clients to connect to specific VLANs/networks. Each peer is assigned to a network via the `vid` field, and their IP address is derived from the network's assignments.
+
+#### Requirements
+
+1. The network must have `assignments` defined (DHCP network with static assignments)
+2. Add a WireGuard peer assignment to the network's assignments
+3. Configure the `wireguard_access` section with peer definitions
+
+#### Configuration
+
+```yml
+wireguard_access:
+  port: 51820                           # WireGuard listen port (optional, default: 51820)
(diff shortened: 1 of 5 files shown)


DEVELOPER.md
```diff
@@ -295,6 +295,105 @@ An optional configuration parameter can be added to any non-tunnel interface to
...
+Then configure the `wireguard_access` section:
+
+```yml
+wireguard_access:
+  port: 51820
+  private_key_file: /root/wireguard.key
+  peers:
+    - assignment: example-laptop-1
+      description: "laptop home network"
+      vid: 40
```
PolynomialDivision, line 351: i think i would prefer here the network name (network.name).

DEVELOPER.md
```diff
@@ -295,6 +295,105 @@ An optional configuration parameter can be added to any non-tunnel interface to
...
+
+wireguard_access:
+  peers:
+    - assignment: example-laptop-1
+      vid: 40
+      public_key: "..."
+    - assignment: example-laptop-2
+      vid: 40
+      public_key: "..."
+```
```
PolynomialDivision, line 383: For me, it looks like the information is duplicated. So the VID is implicitly provided by the DHCP assignment, or can I use one peer in multiple networks (VLANs)?

DEVELOPER.md
Noki: network.name is optional for most locations the dhcp network does not have one. should we make it mandatory?

DEVELOPER.md
Noki: Sounds like a plan. A host could however exist in multiple VLANs so we should enforce that a peer has a single match and raise an exception if not.

PR #1604, opened 2026-04-22
imagebuilder: add kernel debugging as a feature

Noki:


PolynomialDivision:
Why draft?

Noki:
> Why draft?

Did not work. Would require to build the kernel and not use the one comming with the imagebuilder.

Issue #1403, opened 2025-10-23
Explain role ext in developers.md


ZuquetuPlexis:
From time to time i hear of the role **ext**. This role seems to be unexplained in the developers.md or anywhere else. Some explanation/clarification can be found when searching through the matrix channel. I think it should be explained here.

ZuquetuPlexis:
@Noki  Thank you for the explanation. I'm creating a PR based on this.

pktpls:
> @pktpls can you describe why your pod setup uses ext in this way and how it works?

This is a network managed by Podman on the corerouter. Podman (like Docker) creates a bridge over the veth interfaces of the containers. The individual IPs are assigned to the container when it is started: https://github.com/pktpls/bbb-configs/blob/52517099f3f9318f60575b3ecccd207490770dd3/locations/podtwo.yml#L61

For Podman alone it's not neccessary to define this `ext` network, but I want the containers to be routable in the mesh. Example: `mtr mirror.freifunk.dev`

Issue #385, opened 2022-12-03
image is incompatible for sysupgrade


pktpls:
> $ ssh root@huette15-ap1.olsr sysupgrade -n /tmp/huette15-ap1.bin
> Sat Dec  3 23:32:08 CET 2022 upgrade: The device is supported, but this image is incompatible for sysupgrade based on the image version (1.0->9.9).
> Image check failed.

Same without `-n` - I assume this is related to d9d9869, maybe because the EAP225-Outdoor doesn't do switch stuff, neither DSA nor swconfig?

Noki:
This is kind of intended, however the actions to take before upgrading need some documentation. Before upgrading the compat_version should be set to 9.9 using the following:

```
uci set system.@system[0].compat_version="1.1"
uci commit system
```

After that upgrading is possible without problems. This documentation should have been included in the pull request.

PR #1605, opened 2026-04-22
Documentation of so far undocumented features

### LINKED ISSUE #339: documentation: add information about how to add packages to individual location
Instead of adding utils to all locations as proposed in #47 we should add documentation how to install packages for individual location.

Noki:
This PR closes #339, #385 and  #1403.


FFHener (approved):
I am wondering if we need to restructure DEVELOPER.md eventually because it is getting kind of long, but this shouldnt be a blocker for this PR. Thank you for adding more docs

PR #1608, opened 2026-04-23
model: add xiaomi aiot ac2350 v1

ZuquetuPlexis:


group_vars/model_xiaomi_aiot_ac2350.yml
```diff
@@ -0,0 +1,22 @@
+---
+target: ath79/generic
+brand_nice: Xiaomi
+model_nice: AIoT AC2350
+version_nice: v1
+openwrt_version: 25.12-SNAPSHOT
+
+dsa_ports:
+  - eth0.1 # lan Port 2 - 4, Lan 1 - 3
```
Noki: What does this comment mean? I don't understand it.

ZuquetuPlexis: As far as i understand it, their is no way to single out the lan ports. Maybe i did not understand the board.json right.

The OpenWrt website describes the network interface of this device like [this](https://openwrt.org/toh/xiaomi/aiot_router_ac2350?s[]=wifi#network_interfaces)

Noki: ok, so remove the "lan Port 2 - 4" part. ;-)

please also squash into a singe commit.

ZuquetuPlexis: same thing for eth0.2: removing the comment ?

Noki: I would go with:

```
dsa_ports:
  - eth0.1 # LAN 1-3
  - eth0.2 # WAN
```

PR #1609, opened 2026-04-25
perle44: init

pmelange:
add location perle44 at perlebergerstr 44

### DIFF
--- a/locations/perle44.yml
+++ b/locations/perle44.yml
@@ -1,0 +1,66 @@
+---
+location: perle44
+location_nice: Perlebergerstr 44
+latitude: 52.529823
+longitude: 13.345715
+altitude: 36
+contact_nickname: Perle44
+contacts:
+  - Gotthard@Schulte-Tigges.net
+
+hosts:
+  - hostname: perle44-core
+    role: corerouter
+    model: cudy_wr3000e-v1
+    wireless_profile: freifunk_default
+    host__rclocal__to_merge:
+      - |
+        uci -q get system.@system[0].bbb_config_first_boot > /dev/null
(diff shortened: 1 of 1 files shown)


locations/perle44.yml
```diff
@@ -0,0 +1,66 @@
...
+    model: cudy_wr3000e-v1
+    wireless_profile: freifunk_default
+    host__rclocal__to_merge:
+      - |
+        uci -q get system.@system[0].bbb_config_first_boot > /dev/null
+        if [ "1" -eq $? ]; then
+          # set up the WAN port as an untagged connection to the uplink AP
+          uci set network.vlan_10.ports='wan:u'
+          # set up the other ports as untagged for all LAN ports
+          uci set network.vlan_40.ports='lan1:u lan2:u lan3:u lan4:u'
```
PolynomialDivision, line 23: you already set vlan 40 to untagged: true. You should not be able to set wan:u to vlan 10. And the vlan 40 untagged ports are redundant?

pmelange: These are set calls to uci.  I could have used del_list and add_list, but that would be 3 lines.

```
uci del_list network.vlan_10.ports='wan:t'
uci add_list network.vlan_10.ports='wan:u'
uci del_list network.vlan_40.ports='wan:u'
```
and I don't care if vlan 10 is not accessible on the LAN ports.

If there is a better way to have the WAN port untagged for the uplink, and all the LAN ports untagged for the client network, please let me know.

Noki (approved):
can't find the location in hopglass, besides that I think this is good to go. feel free to merge once the location is installed.

PR #1613, opened 2026-05-02
[RFC] adding bird-only-mode

PolynomialDivision:
This will cut the connection between bbb-bird-only-nodes with olsr-only nodes.

### DIFF
--- a/group_vars/all/general.yml
+++ b/group_vars/all/general.yml
@@ -43,6 +43,9 @@ all__sysctl__to_merge:
   vm.min_free_kbytes: 1024
 
 ## Routing Section:
+# Set to false to re-enable OLSR (legacy, for connecting to Falter routers)
+bird_only: true
+
 freifunk_global_prefix: 2001:bf7::/32
 
 # Default mesh metric in inbound direction (rxcost) for normal mesh links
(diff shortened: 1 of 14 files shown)


FFHener:
I would like to discuss a good timeline for this transitionphase at https://github.com/freifunk-berlin/meta/issues/204. In my opion should we proceed with this step only when we switch to babel on falter as well, so people have a upgrade path that keeps working

pktpls:
I like it - we can continue to run OLSR where it's needed, and turn it off everywhere else. But yes let's talk on the 24th.

PolynomialDivision:
What was said about it?

pktpls:
> What was said about it?

I forgot :D But we should do it, I'll take a look at the diff today

pktpls:
I think this is a good approach :+1: It would be useful to know how many Falter nodes there still are wifi-meshing to bbb-configs nodes. Do we need to figure out how these Falter nodes still get at least a default route (and return)? I'm scared of creating lots of little non-routable OLSR islands :)

What about always announcing a default route if we run OLSR? I assume this would suck traffic towards the Babel network, which would be good?

Technical comments about the code:
- make it just `if not bird_only` - it's reasonable to expect the variable to always be present, and it's okay to crash if it's not
- default false probably?

PolynomialDivision:
Superseeded by https://github.com/freifunk-berlin/bbb-configs/pull/1698/commits

PR #1630, opened 2026-05-11
Apup

PolynomialDivision:


FFHener:
Is this compatible with other mesh types? Why are you proposing to use apup?

PolynomialDivision:
Its faster and often better supported and less bugs.

PR #1637, opened 2026-05-16
nhu:init

xa-y-ax:
- [x] deploy devices
- [x] finish config


FFHener:
Well, i think the sxtsq5ac is from zwingli :D Thats why the duplicate-check fails
I will patch it out

PR #1642, opened 2026-05-23
mdk: init

FFHener:


Noki (approved):
looks good to me, but why don't we activate 802.11s mesh here?

FFHener:
The location of the router is unlikely to get mesh neighbours, so I would merge it for the moment

PR #1643, opened 2026-05-23
hacrafu-capelvenere: upgrade core; update vlan, wifi-channel

ZuquetuPlexis:
add ap, static-leases, switch-mgmt, ssh-key

### DIFF
--- a/locations/hacrafu-capelvenere.yml
+++ b/locations/hacrafu-capelvenere.yml
@@ -10,15 +10,18 @@ contacts:
   - "freifunk@hacrafu.de"
 
 hosts:
-
   - hostname: hacrafu-capelvenere
     role: corerouter
-    model: "cudy_wr3000-v1"
+    model: "cudy_wr3000h-v1"
     wireless_profile: freifunk_hacrafu
-    # wifi_roaming: true
     host__rclocal__to_merge:
       - |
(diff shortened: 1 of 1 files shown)


locations/hacrafu-capelvenere.yml
```diff
@@ -146,21 +155,25 @@ networks:
...
 
 # AP-id, wifi-channel, bandwidth, txpower
 location__channel_assignments_11g_standard__to_merge:
-  hacrafu-capelvenere: 13-40
-  hacrafu-capelvenere-dap: 13-40
-  hacrafu-capelvenere-dungeon: 13-20
+  hacrafu-capelvenere: 13-20
+  hacrafu-capelvenere-dap: 9-20
+  hacrafu-capelvenere-range: 13-20
+  hacrafu-capelvenere-dungeon: 9-20
```
FFHener: I would recommend to use non overlapping channels like 1, 6 and 13.

9 and 13 are overlapping.

FFHener (commented):
I would change these channels and squash the 2 commits together.

locations/hacrafu-capelvenere.yml
ZuquetuPlexis: Ok. Please use channel 1 instead of 9

PR #1644, opened 2026-05-24
Bump versions

FFHener:
I upgraded ansible and the linter.

- There is one new deprecation, but we can fix this until the next major version upgrade
- I applied the autofixes from the linter, so the local linting should be more useable again
- Found a little typo in k9


FFHener:
Thats ironic :smile: 
screenshot

PR #1648, opened 2026-05-25
workflows: remove stale branches

Noki:
During the Community Day, we discussed our intention to delete stale branches. This pull request sets up a workflow to accomplish this, with the exception of the `main` branch and any branches that have an associated pull request. We are initially implementing this action in dry-run mode to allow us to review the output; once we are satisfied with the results, we will remove the dry-run flag.

### DIFF
--- a/.github/workflows/remove-stale-branches.yml
+++ b/.github/workflows/remove-stale-branches.yml
@@ -1,0 +1,33 @@
+---
+name: Remove Stale Branches
+
+#
+# Documentation:
+# https://github.com/marketplace/actions/remove-stale-branches
+#
+
+on:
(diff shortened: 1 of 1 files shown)


FFHener:
Thanks for this quick PR. I would suggest rising the default value for stale branches from 90 to something like 180 or 360 and only make it more aggressive if needed just to avoid deleting branches in the future that are still needed (somehow)

Noki:
@FFHener changed to 180 days and 14 days before deletion. The user will have 14 days time after getting the stale notification before it will actually be deleted.

FFHener (approved):

PR #1653, opened 2026-05-28
Images automation

### LINKED ISSUE #1615: Add Github Action for building images if needed
If people arent building images regularly it is not always trivial to build images for them. Therefor we should add a action that builds images. We did this in the pastence for all locations.

I think a Action that only runs on a PR when a certain label is added would be a good solution.

FFHener:
Add a automation that gets triggered by the label `build_image`. This should help people to update routers even if they dont have a working local development setup for building images.

It will detect changed locations and just builds images for them. After that a download link will be posted.  If the user want to have new images after several changes the label would need to be removed and added again. This action is purely optional and wont run by default.

This fixes #1615 

[I tested this on my fork.](https://github.com/FFHener/bbb-configs/pull/28#issuecomment-4567491284)

EDIT: Im not sure if this should be an official way for building the images that we name in the Readme? For now i would leave it as an experimental feature and make sure its usefull the way it is

### DIFF
--- a/.github/workflows/imagerun.yml
+++ b/.github/workflows/imagerun.yml
@@ -1,0 +1,90 @@
+---
+name: Image Run
+
+on:
+  pull_request:
+    branches:
+      - "main"
+    types:
+      - "labeled"
+
+permissions:
+  contents: write # required for artifacts
+  actions: read
+
+jobs:
+  build:
+    name: Ansible Image Run
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Noki (dismissed):
I think this still needs some documentation. I would add a comment block in the workflow yaml file at the top. I'm fine with not having it in the README or DEVELOPER documentation, but it should be easier to understand what the workflow does.

besides the documentation, this looks good to me and can be merged.

PR #1659, opened 2026-06-05
fix openwrt spelling in several files

ZuquetuPlexis:


pktpls:
Nope :)

image

image

ZuquetuPlexis:
> Nope :)
> image image

I'm getting conflicting infos here: @Noki & @pktpls 

Is it OpenWRT or OpenWrt ?

FFHener:
The official branding repository uses camel case as well, so OpenWrt is right: https://github.com/openwrt/branding

Noki:
Please go with OpenWrt as in the brand guide.

FFHener:
Could you drop the first commit? Or squash them together? Besides that it lgtm

PR #1666, opened 2026-06-07
roles: add back freifunk_release file that was lost when removing fal…

Noki:
…ter-common with the difference of using the bbbb-configs hash instead of the falter-packages hash

we need this so the nodes show up with the proper firmware version on the map, otherwise new builds show OpenWrt xx.xx.x instead of Freifunk Falter x.x.x-snapshot


PolynomialDivision:
Nice! Thanks!

PR #1677, opened 2026-06-29
temlpates: add adblock-fast to core routers

PolynomialDivision:
Needs: https://github.com/openwrt/packages/pull/29857

### DIFF
--- a/group_vars/role_corerouter/general.yml
+++ b/group_vars/role_corerouter/general.yml
@@ -2,5 +2,19 @@
 
 wireless_profile: disable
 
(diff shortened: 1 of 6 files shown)


pktpls:
Nice feature, LGTM :+1:

PolynomialDivision:
@pktpls Thanks for merging. :) However, still waiting for this to be merged:
https://github.com/mossdef-org/adblock-fast/pull/19

But it will not break anything, however, without it, it will not work. But looks like it gets soon merged.

PR #1680, opened 2026-07-02
muggel: tweaks for the 5G uplink

pktpls:
Right now the modem loses connectivity every 24h and MBIM commands hang.

With an ifdown+ifup, it recovers after roughly 5 minutes and reconnects. If the connection is alive and healthy, the ifdown+ifup takes just a few seconds.

Let's see if this helps reliably.

### DIFF
--- a/locations/muggel.yml
+++ b/locations/muggel.yml
@@ -19,7 +19,14 @@ hosts:
   - hostname: muggel-core
     role: corerouter
     model: cudy_wr3000e-v1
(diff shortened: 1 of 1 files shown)


pktpls:
Oops I forgot two more commits

- Perf tweaks (hw_offload, wifi channels, tunnel MTU)
- VLAN port config
- More info for replicating the 5G uplink setup

FFHener (approved):

PR #1692, opened 2026-07-07
makefile: add fmt rule

pktpls:
Also:
- update our python dependencies
- fix muggel line length exceedance


Noki (approved):
looks good to me

PR #1698, opened 2026-07-08
locations: add otto, a standalone VDSL/PPPoE site with no mesh

PolynomialDivision:
Adds PPPoE/VDSL uplink support, direct_internet/freifunk_access network flags, uplink_mode: none for tunspace, a zone_wan firewall zone with proper input reject, and a dhcp.j2 host-name collision fix.

direct_internet+freifunk_access networks get a second IPv6 prefix from our own ipv6_prefix (reachable over the mesh) alongside the ISP-delegated one used for internet access, with a SADR route in bird ensuring only the ISP-delegated source can reach the internet exit. babel also no longer announces direct_internet networks that lack freifunk_access, since the firewall wouldn't route them anyway.

Also fixes mandatory()'s ternary-undefined trick silently no-op'ing under ansible-core 2.19, used by both the new uplink validation and the pre-existing getIfname() length check.


PolynomialDivision:
Needs: https://github.com/freifunk-berlin/falter-packages/pull/558

PR #1705, opened 2026-07-10
Sonnenblume vlan ports

Koltonowski:
This adds optional per-network DSA port mappings.

Currently, DSA VLAN generation applies every VLAN to all `dsa_ports`. For devices such as here for example the AVM FRITZ!Box 7530, this means that a network with `untagged: true` is rendered as untagged on all LAN ports.

This change keeps the existing default behavior unchanged:
if no `ports:` field is set, the template still uses all `dsa_ports`.

If `ports:` is set on a network, the template uses that explicit list instead. This allows location configs to express DSA layouts such as:

- VLAN 40: `lan1:t lan2:u lan3:u lan4:u`
- VLAN 50: `lan1:u`

The second line also accepts already annotated entries like `lan1:t` / `lan2:u`, while preserving the old automatic `:t` / `:u` behavior for plain port names.

### DIFF
--- a/locations/sav-hq.yml
+++ b/locations/sav-hq.yml
@@ -1,0 +1,72 @@
+---
+location: sav-hq
+location_nice: Rotherstraße 16, 10245 Berlin
+latitude: 52.5040166
+longitude: 13.4490787
+altitude: 89
+contact_nickname: "Jammingblub"
+contacts:
+  - "freifunk@sva.de"
+
+hosts:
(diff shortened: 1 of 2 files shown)


FFHener (changes requested):
Thank you for trying to implement this feature. Unfortunately this approach doesnt work if you have different devices, e.g. a different Router and AP, as they would have different ports.

And it should fail or provide a way for swconfig devices

PR #1706, opened 2026-07-10
Dsa per network ports

Koltonowski:
This adds optional per-network DSA port mappings.

Currently, DSA VLAN generation applies every VLAN to all `dsa_ports`.  This means that a network with `untagged: true` is rendered as untagged on all LAN ports.

This change keeps the existing default behavior unchanged:
if no `ports:` field is set, the template still uses all `dsa_ports`.

If `ports:` is set on a network, the template uses that explicit list instead. This allows location configs to express DSA layouts such as:

- VLAN 40: `lan1:t lan2:u lan3:u lan4:u`
- VLAN 50: `lan1:u`

The second line also accepts already annotated entries like `lan1:t` / `lan2:u`, while preserving the old automatic `:t` / `:u` behavior for plain port names.

### DIFF
--- a/locations/sav-hq.yml
+++ b/locations/sav-hq.yml
@@ -1,0 +1,72 @@
+---
+location: sav-hq
+location_nice: Rotherstraße 16, 10245 Berlin
+latitude: 52.5040166
+longitude: 13.4490787
+altitude: 89
+contact_nickname: "Jammingblub"
+contacts:
+  - "freifunk@sva.de"
+
+hosts:
+  - hostname: sav-hq # VM on Proxmox
+    role: corerouter
+    model: "x86-64"
+    image_search_pattern: "*-ext4-combined.img*"
+    host__packages__to_merge:
+      - qemu-ga
+
+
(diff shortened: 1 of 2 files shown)


FFHener (changes requested):
See https://github.com/freifunk-berlin/bbb-configs/pull/1705#pullrequestreview-4675604900

pmelange:
There are no checks to see if the list of ports are valid ports.
- tested using ```ports: [foo:u bar:t]```

There are no checks to prevent two vlans having the same port untagged.

This pr makes the ports tag free-form.  The current workaround, the ```host__rclocal__to_merge``` hack, has exactly the same problem.  

I didn't test any swconfig setups list https://github.com/freifunk-berlin/bbb-configs/pull/1705#pullrequestreview-4675604900 mentioned.

pmelange (changes requested):
see https://github.com/freifunk-berlin/bbb-configs/pull/1705#pullrequestreview-4675604900

pmelange:
Also, you need the squash those three commits into one.

PR #1709, opened 2026-07-18
treewide: make fmt

pktpls:
Fixes #1299

### DIFF
--- a/locations/basta.yml
+++ b/locations/basta.yml
@@ -1,8 +1,8 @@
 ---
 location: basta
 location_nice: Schererstr 8
-latitude: 52.548189957239830
-longitude: 13.367031291128984
+latitude: 52.548189957239828
+longitude: 13.367031291128985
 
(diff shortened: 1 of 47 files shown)


locations/basta.yml
Noki, line 1: why does this change?

locations/k9.yml
Noki, line 1: here the second `:` should be removed instead

locations/scherer8-abb.yml
Noki, line 1: why does this change?

locations/k9.yml
pktpls: Haha this one is funny, will fix

locations/basta.yml
pktpls: Something about float precision differences between yaml and python, and how python renders numbers, I would guess

Anyway we don't need locations to be this precise - 7 decimals is already ~1cm precision.

PR #1718, opened 2026-08-11
model: xiaomi aiot ac2350: use swconfig and openwrt 25.12.5

ZuquetuPlexis:


FFHener:
Could you shortly explain the switch from dsa to swconfig?

ZuquetuPlexis:
> Could you shortly explain the switch from dsa to swconfig?

The wireguard tunnels did not connect to their uplink interface. This does not happen when using swconfig.

I could not figure out why the uplink tunnel did not work.

PR #1722, opened 2026-08-15
templates: bgpdisco/dnsmasq: don't use SIGHUP

PolynomialDivision:


Noki:
tested and works for me

PR #1749, opened 2026-09-18
freifunk_release: take this file more seriously

pktpls:
- Set our own DISTRIB_ID, bbb-configs is not Falter.
- Set VARIANT based on the host role.
- Set VARIANT to "tunnel" if at least one tunnel network is defined.
- Move the external tool call for REVISION to the toplevel playbook.
- Add the file to ap hosts as well.


pktpls:
Example:
```
FREIFUNK_DISTRIB_ID="bbb-configs"
FREIFUNK_RELEASE="1.6.0-snapshot"
FREIFUNK_REVISION="32357459"
FREIFUNK_VARIANT="tunnel" # or ap/corerouter/gateway
```

